PR #6, opened 2023-06-30
New output format

Jazarro:
ANSI support:
- Added a dependency on the ansi_term crate.
- The ErrorLevel and the ErrorKey are currently the only elements that are coloured.
- The ck3-tiger.conf file now has an output_style block that allows setting colours for each error level. It also allows disabling styled output altogether.
- The program accepts a --no-colour flag that totally disables ANSI characters in the output. It overrides whatever is in the config.

Minor fixes:
- Added a newline in the output between the preamble and the first error report.
- The --pod flag had a duplicated comment.


src/main.rs
```diff
@@ -186,10 +192,19 @@ fn main() -> Result<()> {
...
     }
     eprintln!("Using mod directory: {}", modpath.display());
+    eprintln!();
+
     set_mod_root(modpath.clone());
 
     let mut everything = Everything::new(&args.ck3.unwrap(), &modpath, modfile.replace_paths())?;
     everything.load_all();
+
+    // We must apply the --no-colour flag AFTER loading and applying the config,
```
amtep, line 202: I'm afraid this won't work the way you expect, because everything.load_all() already emits parse errors and such before validation

src/main.rs
```diff
@@ -113,6 +116,9 @@ fn find_ck3_directory() -> Option<PathBuf> {
 }
 
 fn main() -> Result<()> {
+    #[cfg(windows)]
+    let _ = ansi_term::enable_ansi_support();
```
amtep, line 120: I tried to look up this enable_ansi_support function to see why the return value was being thrown away, but I couldn't find it in the docs. Do you have a reference?

amtep (changes requested):

src/everything.rs
```diff
@@ -336,6 +339,26 @@ impl Everything {
...
+    ///
+    /// Note that the settings from the config can still be overridden
+    /// by supplying the --no-colour flag.
+    fn load_output_styles(&self) -> Option<OutputStyle> {
+        let block = self.config.get_field_block("output_style")?;
+        if !block.get_field_bool("enable").unwrap_or(true) {
+            return Some(OutputStyle::no_colour());
+        }
+        let mut style = OutputStyle::default();
+        if let Some(error_block) = block.get_field_block("error") {
```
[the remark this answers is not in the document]
Jazarro: Ah, I meant to do those, guess I forgot.

src/main.rs
Jazarro: It returns an Error, I guess unwrapping it would be nicer.

[LINK TO CODE](https://github.com/ogham/rust-ansi-term/blob/ff7eba98d55ad609c7fcc8c7bb0859b37c7545cc/src/windows.rs#L10)

src/errors.rs
```diff
@@ -242,9 +247,20 @@ impl Errors {
...
+            self.styles.style(&Styled::Tag(level, false)).paint("("),
+            self.styles
+                .style(&Styled::Tag(level, false))
+                .paint(format!("{}", key)),
+            self.styles.style(&Styled::Tag(level, false)).paint(")"),
+        ];
         writeln!(
             self.outfile.as_mut().expect("outfile"),
-            "{level} ({key}): {msg}"
+            "[{}]: {msg}",
```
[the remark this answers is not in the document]
Jazarro: It looks better to me, but that's entirely subjective of course.

src/main.rs
```diff
@@ -124,6 +127,12 @@ fn find_ck3_dir_in_steam_dir(steam_dir: PathBuf) -> Option<PathBuf> {
 fn main() -> Result<()> {
     let mut args = Cli::parse();
 
+    #[cfg(windows)]
+    if !args.no_color {
+        let _ = ansi_term::enable_ansi_support()
+            .map_err(|_| eprintln!("Failed to run enable_ansi_support() for Windows10 users. Continuing probably without colored output."));
```
amtep, line 133: Since this is user-facing it should probably just start with "Failed to enable ANSI support ..." instead of naming the function.

src/everything.rs
```diff
@@ -336,6 +340,30 @@ impl Everything {
         }
     }
 
+    /// Load the OutputStyle settings from the config.
+    /// Returns None if no settings are defined.
+    /// Otherwise, returns the overwritten OutputStyles.
+    ///
+    /// Note that the settings from the config can still be overridden
+    /// by supplying the --no_color flag.
```
amtep, line 348: Very minor, but it's actually the --no-color flag :) (dash not underscore)

src/errors.rs
```diff
@@ -206,6 +214,229 @@ impl Errors {
...
+            self.styles.style(&Styled::Default).paint(" "),
+            self.styles
+                .style(&Styled::Tag(severity, true))
+                .paint(format!(
+                    "{}",
+                    pointer
+                        .msg
+                        .as_ref()
+                        .map(|_| "<-- ")
+                        .unwrap_or(&"".to_string())
```
amtep, line 419: Can &"".to_string() just be "" ?

amtep (approved):

src/errors.rs
```diff
@@ -249,18 +245,31 @@ impl Errors {
         // Log the primary pointer:
         self.log_pointer_primary(report.primary(), report.indentation(), report.lvl.severity);
         // Log the other pointers:
-        report.pointers.iter().skip(1).for_each(|pointer| {
+        report.pointers.iter().enumerate().skip(1).for_each(|(index, pointer)| {
```
amtep, line 248: This might be more idiomatic with slice.windows() :)
https://doc.rust-lang.org/std/primitive.slice.html#method.windows

src/helpers.rs
```diff
@@ -50,7 +50,7 @@ pub fn dup_assign_error(key: &Token, other: &Token) {
     log(LogReport {
         lvl: LogLevel::new(Severity::Warning, Confidence::Reasonable),
         key: ErrorKey::DuplicateField,
-        msg: format!("`{other}` is redefined in a following line"),
+        msg: format!("`{other}` is redefined in a following line").as_str(),
```
amtep, line 53: I think you can just do `&format!`

src/errors.rs
Jazarro: Amazing, I had no idea this existed.

ck3-tiger/src/main.rs
```diff
@@ -192,9 +201,17 @@ fn main() -> Result<()> {
         bail!("Cannot find mod directory. Please make sure the .mod file is correct.");
     }
     eprintln!("Using mod directory: {}", modpath.display());
+    eprintln!();
```
amtep, line 204: I think this has to be moved down a bit, because the Everything module does eprintln when loading extra mods. This happens while loading the config file.

src/block.rs
```diff
@@ -5,9 +5,9 @@ use std::str::FromStr;
 
 pub mod validator;
 
-use crate::errorkey::ErrorKey;
-use crate::errors::{error, error_info};
 use crate::parse::pdxfile::{parse_pdx_macro, LocalMacros};
+
```
amtep, line 9: These extra blank lines will be a problem because then `cargo fmt` won't nicely alphabetize the imports anymore :)

src/data/scripted_effects.rs
```diff
@@ -166,7 +166,7 @@ impl Effect {
         tooltipped: Tooltipped,
     ) {
         // Every invocation is treated as different even if the args are the same,
-        // because we want to point to the correct one when reporting errors.
+        // because we want to point to the correct one when report errors.
```
amtep, line 169: Overactive regexp, I think. I found several of these

src/lib.rs
```diff
@@ -42,6 +39,7 @@ mod modif;
 mod output_style;
 mod parse;
 mod pdxfile;
+pub mod report;
```
amtep, line 42: The `pub mod` are in their own section higher up

src/fileset.rs
```diff
@@ -17,6 +17,7 @@ use crate::report::ErrorKey;
 use crate::report::{add_loaded_mod_root, error, warn_abbreviated, warn_header, will_log};
 use crate::token::{Loc, Token};
 
+/// Note that ordering of these enum values matters. Do not change the order.
```
amtep, line 20: Instead of "do not change the order", I'd like to explain a bit more. ("Do not..." comments are a pet peeve of mine :)
Maybe "The files from later ones will override the files from earlier ones, and the error reporting is sensitive to the order when deciding what to log."

src/report/error_key.rs
```diff
@@ -27,6 +27,7 @@ pub enum ErrorKey {
     ImageFormat,
     Unneeded,
     Scopes,
+    /// This seems like it solves the same problem that `Confidence` solves.
```
amtep, line 30: Yeah the purpose is to hide the StrictScopes ones behind a command-line flag. This could be the same flag that activates weak-confidence reports.

Basically StrictScopes will warn if you use a scope name that has not been previously defined. But it means I have to catch all places where the game engine predefines scopes. It also gets more easily confused by complicated scripted effects than the normal mode does.

src/report/filter.rs
```diff
@@ -0,0 +1,137 @@
...
+                return false;
+            }
+        }
+        true
+    }
+    /// Check all elements of the loc link chain.
+    /// This is necessary because of cases like a mod passing `CHARACTER = this` to a vanilla
+    /// script effect that does not expect that. The error would be located in the vanilla script
+    /// but would be caused by the mod.
+    fn get_end_of_link_chain(loc: &Loc) -> &Loc {
```
amtep, line 63: Hmm here you just check the top level of the link chain. I don't think that works right.

A common example of blacklist use is if a mod copies vanilla files into the mod, and doesn't want to fix bugs in the copied files. Then the modder adds those vanilla files to the blacklist, and errors located in them should not be reported. Even if the mod called into that file. The modder would not want to add the mod files to the blacklist to cover up these reports, because errors in the modded files need to be caught.

For a similar reason, the no-vanilla check should also check all links in the chain, but with the opposite logic: it should print the report if any of them are modded files. This covers the case where a mod replaces a scripted effect that then causes errors when called from vanilla locations. (Blacklisting should be stronger than this check.)

src/fileset.rs
Jazarro: The enum implements PartialOrd and Ord. Elsewhere in the code, enum values are compared using inequality operators. This means changing the order will fundamentally change application logic, almost definitely resulting in undesirable behaviour. This fact may not be obvious to the passer-by programmer, which is why I added the warning. There was no documentation on it previously. It is not my enum.

src/report/filter.rs
Jazarro: I might be wrong, but I'm pretty sure that's what it always used to do.

Jazarro: Never mind, I see now.

amtep: I realize now that my partially short-circuited recursion was hard to read :) could definitely use a comment

ck3-tiger.conf
```diff
@@ -13,21 +13,66 @@ languages = {
...
-	# If there are no keys, all errors in the specified files will be ignored.
-	key = brace-placement
+# Allows configuring a number of ways to reduce the number of reports that are printed by CK3-Tiger.
+filter = {
 
-	# Each file entry specifies a file in which to ignore the specified key or keys.
-	# If there are no file entries, the keys are ignored in all files.
-	file = events/POD_initialization/POD_maintenance_events.txt
+    # Reports that aren't important enough will not be printed.
+    # Valid values are: Error, Warning, Info, Untidy.
```
amtep, line 24: I've been thinking about those, and it should maybe be

Error, Warning, Untidy, Tips

"Info" is currently used only by advice-level reports and those are used for tips: things that could be improved but aren't wrong. I hope to expand that section actually, especially regarding performance tips.

I think they should be below "untidy" because untidy things are slightly wrong, while tips are about things that are not wrong at all. The Rust equivalent would be clippy suggesting more idiomatic ways to write code.

I've also been using advice for some Untidy things but I can change those when the new system is in.

ck3-tiger.conf
```diff
@@ -13,21 +13,66 @@ languages = {
...
+    # Valid values are: Weak, Reasonable, Strong.
+    # This is an optional value, it defaults to Reasonable.
+    minimum_confidence = Reasonable
+
+    # Whether to report about problems in vanilla game files.
+    # Setting this to 'yes' results in a LOT of spam.
+    # Optional boolean value, defaults to no.
+    show_vanilla = no
+
+    # Whether to report about problems in other loaded mods.
```
amtep, line 38: "... in mods loaded via the load_mod sections" is probably clearer

ck3-tiger.conf
```diff
@@ -65,25 +110,30 @@ output_style = {
     # Also, all other settings in this block will be ignored.
     enable = yes
 
+    # The maximum line-length of the output.
+    # Set this to 0 to disable length restrictions.
+    # This is an optional value, it defaults to 120.
+    max_line_length = 120
```
amtep, line 116: In max_line_length also ignored when enable = no?

If not, it could be moved up and the comments for enable can say "the settings below". Or enable can be renamed to enable_colors

src/everything.rs
```diff
@@ -26,9 +27,9 @@ use crate::data::buildings::Building;
 use crate::data::casusbelli::{CasusBelli, CasusBelliGroup};
 use crate::data::character_templates::CharacterTemplate;
 use crate::data::characters::Characters;
-use crate::data::coa::{CoaDynamicDefinition, CoaTemplateList, Coas};
+use crate::data::coa::{CoaDynamicDefinition, Coas, CoaTemplateList};
 use crate::data::coadesigner::{
-    CoaDesignerColorPalette, CoaDesignerColoredEmblem, CoaDesignerEmblemLayout, CoaDesignerPattern,
+    CoaDesignerColoredEmblem, CoaDesignerColorPalette, CoaDesignerEmblemLayout, CoaDesignerPattern,
```
amtep, line 32: Are you using different cargo fmt settings? I'm using the defaults (I think).

src/everything.rs
```diff
@@ -395,8 +495,8 @@ impl Everything {
     /// A helper function for categories of items that follow the usual pattern of
     /// `.txt` files containing a block with definitions
     pub fn load_pdx_items_ext<F>(&mut self, itype: Item, add: F, ext: &str)
-    where
-        F: Fn(&mut Db, Token, Block),
+        where
```
amtep, line 498: This also looks like different rustfmt settings

src/everything.rs
```diff
@@ -305,8 +307,106 @@ impl Everything {
...
+                error(value, ErrorKey::Config, "Missing key. Loose values are not allowed here.");
+                continue;
+            }
+            let key = key_opt.as_ref().expect("Should exist.");
+            if comp != &Eq {
+                error(key, ErrorKey::Config, &format!("Detected illegal operator `{comp}`, only `=` is allowed."));
+                continue;
+            }
+            match (key.as_str(), &value) {
+                ("key", _) => error(key, ErrorKey::Config, "Detected illegal key: `key` should probably be `rule_key`."),
```
amtep, line 351: Can the keys be defined without the `rule_` prefix in the first place? To keep the config file simpler

src/everything.rs
```diff
@@ -305,8 +307,106 @@ impl Everything {
...
+
+    fn load_filter_rules_list(&self, rules_type: RulesListType, rules: &Block) {
+        for (key_opt, comp, value) in rules.iter_items() {
+            if let None = key_opt {
+                error(value, ErrorKey::Config, "Missing key. Loose values are not allowed here.");
+                continue;
+            }
+            let key = key_opt.as_ref().expect("Should exist.");
+            if comp != &Eq {
+                error(key, ErrorKey::Config, &format!("Detected illegal operator `{comp}`, only `=` is allowed."));
```
amtep, line 347: I prefer a gentler style of communication with the user :)

Like

Found operator `{comp}`, expected only `=`

Or the ones further down:

Unknown key `key`. Did you mean `rule_key`?

src/everything.rs
Jazarro: Yea, I'm using the default settings, I think. It looks like it's just sorting it alphabetically. Does yours not do that? Maybe I'm on a different version.

PR #10, opened 2023-07-06
Filter configuration

Jazarro:
- Extracts all filter-related code into its own file.
- Completely revamped filter syntax and logic. Filters are now built like triggers, using NOT, OR, etc. This makes it easier to apply more complex filters. Whitelisting keys and files also becomes much easier.
- Added a guide in `filter.md`

### DIFF
--- a/filter.md
+++ b/filter.md
@@ -1,0 +1,223 @@
+# Filtering the output
+
+Tiger generates a lot of reports. This guide explains how to use filters to control what reports are printed.
+
+In the `ck3-tiger.conf` file, add a `filter` block. Inside the `filter`, add a `trigger` block.
+
+```
+# You can have at most 1 filter block.
+filter = {
+    # Whether to report about problems in vanilla game files.
+    # Setting this to 'yes' results in a LOT of spam.
+    # Optional boolean value, defaults to no.
+    show_vanilla = no
+
+    # Whether to report about problems in mods loaded via the load_mod sections.
+    # Optional boolean value, defaults to no.
+    show_loaded_mods = no
+    
+    # You can have at most 1 trigger block. It is an implicit AND trigger, just like in paradox script.
+    trigger = {
+        # Read on to see what triggers can be used here.
+    }
+}
+```
+
+## Triggers
+
+This is the list of valid triggers that can be used in the `trigger` block.
+
+### always
+
+- `always = yes` matches all reports
+- `always = no` matches no reports
+
+### Logic gates
+
+You can use any of the following logic gates.
+
+- `NOT = { <trigger> }`
+    - True if the trigger is false.
+- `AND = { <trigger> <trigger> ... }`
+    - True if all triggers are true.
+    - An empty `AND` block is the same as `always = yes`.
+- `OR = { <trigger> <trigger> ... }`
+    - True if at least one trigger is true.
+    - An empty `OR` block is the same as `always = no`.
+- `NAND = { <trigger> <trigger> ... }`
+    - True unless all triggers are true.
+    - This is the same as writing `NOT = { AND = { } }`
+- `NOR = { <trigger> <trigger> ... }`
+    - True if all triggers are false.
+    - This is the same as writing `NOT = { OR = { } }`
+
+### Severity
+
+Severity is a measure of the seriousness of the problem. For example; using the wrong level of indentation is an
+extremely minor problem that wouldn't impact the functionality of the mod at all. It is a low severity.
+Conversely, a problem that causes the game to crash on start-up is of the highest severity.
+
+These are the severity levels:
+
+- `Tips` These are things that aren't necessarily wrong, but there may be a better, more idiomatic way to do it. This
+  may also include performance issues.
+- `Untidy` This code smells. The player is unlikely to be impacted directly, but developers working on this codebase
+  will likely experience maintenance headaches.
+- `Warning` This will result in glitches that will noticeably impact the player's gaming experience. Missing
+  translations are an example.
+- `Error` This code probably doesn't work as intended. The player may experience bugs.
+- `Fatal` This is likely to cause crashes.
+
+You can add the severity trigger like so:
+
+- `severity = Tips` (Only match tips.)
+- `severity >= Warning` (Match Warnings or higher.)
+- `severity < Error` (Match anything below Error.)
+
+### Confidence
+
+Describes how confident Tiger is that a problem is real and not a false positive. If you are spammed with many false
+positives, you may try raising the required confidence level.
+
+These are the current confidence levels. They are subject to change.
+
+- `Weak` Tiger is not very confident about this report, it's quite likely to be a false positive.
+- `Reasonable` The default level.
+- `Strong` Tiger is very confident that this is a real problem.
+
+You can add the confidence trigger like so:
+
+- `confidence = Weak` (Only match weak.)
+- `confidence >= Reasonable` (Match Reasonable or higher.)
+- `confidence < Strong` (Match anything below Strong.)
+
+### Key
+
+Every report has a `key`, it describes the type of problem that is being reported about. You can find the key in
+parentheses on the first line of the report. For example, in the report below, the key is `missing-localization`:
+
+```
+Error(missing-localization): missing english localization key Badlay
+   --> [CK3] history/characters/afar.txt
+267 |  name = "Badlay"
+    |         ^ 
+```
+
+You can add a key trigger like so:
+
+- `key = missing-localization` This will match reports with the `missing-localization` key.
+
+### File
+
+You can target files and folders. Reports that mention the file will be matched by this trigger.
+
+Example:
+
+- `file = common/` This matches any report that mentions a file inside the `common/` directory.
+- `file = history/characters/afar.txt` This matches any report that mentions that specific file.
+
+# Migrating from `ignore` to `filter`
+
+Filtering was previously done through `ignore` blocks.
+
+Suppose this is our old setup:
+
+```
+ignore = {
+    key = duplicate-field
+}
+ignore = {
+    file = events/travel_events/travel_events.txt
+}
+ignore = {
+    key = duplicate-item
+    file = common/decisions/80_major_decisions.txt
+    file = common/script_values/00_basic_values.txt
+    file = common/defines/00_defines.txt
+}
+```
+
+We can replicate this like so:
+
+```
+trigger = {
+    NOR = {
+        key = duplicate-field
+        file = events/travel_events/travel_events.txt
+    }
+    NAND = {
+        key = duplicate-item
+        OR = {
+            file = common/decisions/80_major_decisions.txt
+            file = common/script_values/00_basic_values.txt
+            file = common/defines/00_defines.txt
+        }
+    }
+}
+```
+
+# Examples
+
+## Filter by severity
+
+```
+trigger = {
+    # Only print Warnings or above:
+    severity >= Warning
+    # Don't print reports that are likely false positives:
+    confidence > Weak
+}
+```
+
+## Whitelist the folder you're working on
+
+```
+trigger = {
+    # Only reports about files inside this folder will be printed.
+    file = common/traits/
+}
+```
+
+## Ignore a certain key
+
+```
+trigger = {
+    # I don't care about localization (yet), I'll add the translations later!!!
+    NOT = { key = missing-localization }
+}
+```
+
+## All of the above
+
+A more realistic case. You're working on implementing custom traits, and want to see warnings and errors specific to
+traits. You're planning on adding translations later, so you temporarily blacklisted that particular key.
+
+```
+trigger = {
+    # Only print warnings and errors:
+    severity >= Warning
+    # Don't print reports that are likely false positives:
+    confidence >= Reasonable
+    # Only reports about files inside this folder will be printed.
+    file = common/traits/
+    # I don't care about localization (yet), I'll add the translations later!!!
+    NOT = { key = missing-localization }
+}
+```
+
+## Custom severity level for spammy file
+
+```
+trigger = {
+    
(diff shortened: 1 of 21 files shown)


filter.md
```diff
@@ -0,0 +1,217 @@
...
+    # Optional boolean value, defaults to no.
+    show_vanilla = no
+
+    # Whether to report about problems in mods loaded via the load_mod sections.
+    # Optional boolean value, defaults to no.
+    show_loaded_mods = no
+    
+    # You can have at most 1 trigger block.
+    trigger = {
+        # The root trigger block is an AND-gate, just like in paradox script.
```
amtep, line 21: Calling it the root trigger block is confusing for modders because "root" means something else. Maybe just say "You can have at most 1 trigger block. It is an implicit AND trigger, just like in paradox script."

filter.md
```diff
@@ -0,0 +1,217 @@
...
+
+- `always = yes` matches all reports
+- `always = no` matches no reports
+
+### Logic gates
+
+You can use any of the following logic gates. The `NOT` gate must contain exactly one trigger. All other gates take zero
+or more triggers.
+
+- `NOT = { <trigger> }`
```
amtep, line 40: If you want it to be like pdx script, NOT should be an implicit NOR :)

filter.md
```diff
@@ -0,0 +1,217 @@
...
+- `AND = { <trigger> <trigger> ... }`
+    - Matches reports that match all children.
+    - An `AND` block without children is the same as `always = yes`.
+- `OR = { <trigger> <trigger> ... }`
+    - Matches reports that match at least one child.
```
amtep, line 45: "True if any of the triggers are true"

filter.md
```diff
@@ -0,0 +1,217 @@
...
+    - An `OR` block without children is the same as `always = no`.
+- `NAND = { <trigger> <trigger> ... }`
+    - This is the same as writing `NOT = { AND = { } }`
+- `NOR = { <trigger> <trigger> ... }`
+    - This is the same as writing `NOT = { OR = { } }`
```
amtep, line 50: True if none of the triggers are true

filter.md
```diff
@@ -0,0 +1,217 @@
...
+This is the list of valid triggers that can be used in the `trigger` block.
+
+### always
+
+- `always = yes` matches all reports
+- `always = no` matches no reports
+
+### Logic gates
+
+You can use any of the following logic gates. The `NOT` gate must contain exactly one trigger. All other gates take zero
```
amtep, line 37: Keep in mind that many modders are not programmers. Talk of logic gates and gates having children is going to confuse them.

filter.md
```diff
@@ -0,0 +1,217 @@
...
+    - Matches reports that match at least one child.
+    - An `OR` block without children is the same as `always = no`.
+- `NAND = { <trigger> <trigger> ... }`
+    - This is the same as writing `NOT = { AND = { } }`
+- `NOR = { <trigger> <trigger> ... }`
+    - This is the same as writing `NOT = { OR = { } }`
+
+### Severity
+
+Severity is a measure of how serious the problem being reported about is. For example; using three spaces of indentation
```
amtep, line 54: Pdx script is indented with tabs :)

filter.md
```diff
@@ -0,0 +1,217 @@
...
+
+We can replicate this like so:
+
+```
+trigger = {
+    NOR = {
+        key = duplicate-field
+        file = events/travel_events/travel_events.txt
+    }
+    NAND = {
```
amtep, line 143: NAND is the hardest to understand so it's a bit of a problem if that's going to be the most common one

src/block.rs
```diff
@@ -552,61 +552,93 @@ impl Block {
...
+    /// Substitutes `==` for `=`.
+    /// This is to allow `=` to be used as an alias for equality,
+    /// even though that's technically incorrect.
+    pub fn to_comparator(self) -> Self {
+        if self == Self::Eq {
+            Self::Equals
+        } else if self.is_comparator() {
+            self
+        } else {
+            panic!("Cannot convert {self} to comparator operator. Check is_comparator() first.");
```
amtep, line 625: I don't like that this can panic. Can we do without?

src/block.rs
```diff
@@ -552,61 +552,93 @@ impl Block {
...
+                | Operator::NotEquals
+                | Operator::AtLeast
+                | Operator::AtMost
+                | Operator::GreaterThan
+                | Operator::SmallerThan
+        )
+    }
+    /// Substitutes `==` for `=`.
+    /// This is to allow `=` to be used as an alias for equality,
+    /// even though that's technically incorrect.
```
amtep, line 618: Sorry but this is a very un-pdx-script attitude. = is the default equality operator, used *everywhere* in over 4 million lines of script, except for four places where == is used.

Please treat = like a normal operator.

src/report/errors.rs
```diff
@@ -130,13 +85,9 @@ impl Errors {
...
-        }
-        if !self.will_log(&report.primary().location, report.key) {
-            return;
         }
-        log_report(self, &report);
+        self.seen.insert(index);
+        log_report(self, report);
     }
 
     pub fn log_abbreviated(&mut self, loc: &Loc, key: ErrorKey) {
```
amtep, line 93: This and push_header do need their will_log checked, otherwise users will get a long list of "unused" logs that they didn't ask for

src/block.rs
Jazarro: This does convert `=` to `==`, so both are handled the same. 

Basically, the `to_comparator()` function can be called when the operator is expected to be a comparator, and you'll know that it'll be an Equals, not an Eq. This way, you don't have to do two checks instead of one later-on in the code. 

Again, this is only in tiger's codebase. As far as pdx script is concerned, both are treated the same and no warnings are issued to the user.

src/block.rs
Jazarro: The function was meant to be executed after you've already checked the is_comparator() function, so if used that way it can't ever panic. 

There might be a better way of doing it, maybe returning an option or something, but then you'd have to expect() that and you're right back to a potential panic.

filter.md
Jazarro: I'll see if I can rewrite the guide for normies. :)

filter.md
Jazarro: I could add syntactic sugar for that one. Something like 
```
keys_in_files = {
    keys = {
         key1 key2 key3
    }
    files = {
         file2 file2 file3
    }
}
```

src/report/errors.rs
Jazarro: Ah yes, this is a slight regression. Although, as I understand it, that whole reporting method is locked behind a CLI flag anyways.

I was planning on seeing if I could bring this method of logging more in line with the standard reports.

filter.md
Jazarro: I didn't know that.

src/report/errors.rs
amtep: Ah right. The CLI flag used to use an ignore-key to toggle the behavior :) But that got changed. So it's less important now.

amtep: The density of this method is the main thing. One per line. And easy to extract just the filenames from the unused files one.

src/block.rs
amtep: I still find this logic strange because they're all comparators (except the None one) and you renamed them to Operator for mysterious reasons :) I would just handle all of them at the call site, instead of using these functions to filter out one or two of them.

amtep: I'll probably get rid of the None one by making Block contain an enum-vec instead of a tuple-vec, but that's a big change that will touch all the code.

filter.md
amtep: That would be good :)

src/block.rs
amtep: Okay but even within tiger's code base, = should be treated as the normal equality operator and == as the weird exception, otherwise you get an impedance mismatch with the code it's scrutinizing.

Examples should also use =.

filter.md
```diff
@@ -0,0 +1,217 @@
...
+
+- `Tips` These are things that aren't necessarily wrong, but there may be a better, more idiomatic way to do it. This may also include performance issues.
+- `Untidy` This code smells. The player is unlikely to be impacted directly, but developers working on this codebase will likely experience maintenance headaches.
+- `Warning` This will result in glitches that will noticeably impact the player's gaming experience. Missing translations are an example.
+- `Error` This code probably doesn't work as intended. The player may experience bugs.
+- `Fatal` This is likely to cause crashes.
+
+You can add the severity trigger like so:
+
+- `severity == Tips` (Only match tips.)
```
amtep, line 68: Here's where you're using == in examples

filter.md
```diff
@@ -0,0 +1,217 @@
...
+
+These are the current confidence levels. They are subject to change.
+
+- `Weak` Tiger is not very confident about this report, it's quite likely to be a false positive.
+- `Reasonable` The default level.
+- `Strong` Tiger is very confident that this is a real problem.
+
+You can add the confidence trigger like so:
+
+- `confidence == Weak` (Only match weak.)
```
amtep, line 85: Here too.

The idea was that triggers would be more familiar to modders, right? Most modders won't even know that == is a thing in pdx script.

src/block.rs
```diff
@@ -552,61 +552,93 @@ impl Block {
...
-    /// The == operator, which means Eq but cannot be used to assign
-    EEq,
-    /// The ?= operator
-    QEq,
-    Lt,
-    Gt,
-    Le,
-    Ge,
-    Ne,
+    /// The ?= operator Can be used to open a scope on the condition that it exists.
```
amtep, line 560: ?= can also be used as a conditional comparator. `scope:faith ?= faith:catholicism` is an equality comparison with a short-circuit to false if scope:faith does not exist. The main difference with = is that it won't log an error. Unfortunately, modders have been enthusiastically using ?= everywhere to hide their errors.

src/block.rs
```diff
@@ -552,61 +552,93 @@ impl Block {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub enum Comparator {
+pub enum Operator {
+    /// Can perhaps be replaced by wrapping the operator in an option.
     None,
-    /// Eq is also Assign
+    /// The = operator. Can be used as assignment, or to open a scope.
```
amtep, line 558: This is what I mean by impedance mismatch. This comment should say that Eq is the equality operator. And calling the one for == "Equals" is just confusing because it's so rarely used and calling it "Equals" makes it step on the domain of Eq.

I don't think these new names are an improvement. `MaybeEq` and `Equals` are misleading, which is worse than cryptic.

src/block.rs
```diff
@@ -552,61 +552,93 @@ impl Block {
...
-    Ge,
-    Ne,
+    /// The ?= operator Can be used to open a scope on the condition that it exists.
+    MaybeEq,
+    /// ==
+    Equals,
+    /// !=
+    NotEquals,
+    /// <
+    SmallerThan,
```
amtep, line 567: LessThan is the conventional term :)

src/block.rs
```diff
@@ -552,61 +552,93 @@ impl Block {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub enum Comparator {
+pub enum Operator {
```
amtep, line 555: I don't feel this name change is warranted. All of them except None are comparators. Eq and QEq also do other things, but `Comparator` is a good way to describe them (and therefore vague)

src/block.rs
Jazarro: If you really want to change it back to Operator we can do that. 

My thinking was thus: they are all operators, so that name is definitely correct. Most are used to compare ordered values, but the assignment operator, None and the question mark operators are not. Therefore, Comparator is a somewhat misleading name.

I don't like the panic there either, I'll see if something can be done.

filter.md
Jazarro: Is that a problem? Both are supported. Might as well use the more correct one in the examples. Unless you really hate it, I can change it.

filter.md
Jazarro: If we're going to treat them exactly the same, then why not drop the `==` version from the enum?

src/block.rs
Jazarro: " This comment should say that Eq is the equality operator. " But it isn't the equality operator. That enum value doesn't represent that. Unless you want to make it represent both equality and assignment. In that case, we might as well get right of the other one, the one currently called Equals. There is no point in having two enum values that have overlapping meanings.

As for MaybeEq, why is that misleading? I couldn't think of a different name. Is there an official name for that operator?

src/block.rs
Jazarro: The comments are misleading here, but to my mind, the enum values don't actually have anything to do with notation, they should be viewed as different things. One represents equality comparison, the other assignment and opening scopes. I don't see Operator::Equals as having anything to do with `==` or `=`.

But I think the real problem is having both in the first place. I think we should merge them. Within paradox script, there clearly isn't a real distinction in how they're handled. In fact, `==` seems to be some kind of weird little-known, little-used alias for `=`. We should just serialise both to Comparator::Equals.

src/block.rs
Jazarro: I didn't know ?= could also be used as a conditional equality comparator. That sheds a new light on things, Comparator now makes more sense as a name for the enum. I have changed it.

src/block.rs
Jazarro: As I stated in a different comment: I didn't realise ?= was a comparison operator too. I changed it back. As for the panic: if we merge the two enum values that whole function can be deleted.

src/block.rs
Jazarro: I renamed it `ConditionalEquals` for now. Maybe it could be called `QEquals` instead? While slightly more cryptic, it is a lot shorter.

PR #14, opened 2023-07-29
Added imperator-tiger

dementive:
The imperator-tiger feature flag doens't compile...yet but the other flags still function correctly.


src/imperator/tables/modifs.rs
```diff
@@ -0,0 +1,378 @@
...
+        if warn
+            && !data.item_exists(Item::Terrain, part)
+            && !data.item_exists(Item::Unit, s)
+        {
+            let msg = format!("could not find any {part}");
+            let info = "Could be: $Terrain$_combat_bonus or $Unit$_$Terrain$_combat_bonus";
+            warn_info(name, ErrorKey::MissingItem, &msg, info);
+        }
+        return Some(ModifKinds::Country);
+    }
```
amtep, line 102: This needs to be more complicated... for the `$Unit$_$Terrain$_combat_bonus` variant, you need to match both a unit and a terrain. Vic3 has one like that too, `building_group_$BuildingGroup$_$PopType$_fertility_mult`. You can look there for an example of how to handle this.

It'll be easier if neither unit nor terrain are likely to have `_` in their names.

src/imperator/tables/on_action.rs
```diff
@@ -0,0 +1,246 @@
...
+pub fn on_action_scopecontext(key: &Token, _data: &Everything) -> Option<ScopeContext> {
+    if let Some(oa_sc) = ON_ACTION_SCOPES.get(key.as_str()) {
+        let mut sc = ScopeContext::new(oa_sc.root, key);
+        for (name, s) in &oa_sc.names {
+            sc.define_name(name, *s, key);
+        }
+        for (list, s) in &oa_sc.lists {
+            sc.define_list(list, *s, key);
+        }
+        sc.set_strict_scopes(false);
```
amtep, line 27: Lol this line is a bug, and it's a bug in the other two games as well. Should be deleted; this `sc` can be strict because the defined scope names are known.

src/imperator/tables/effects.rs
```diff
@@ -0,0 +1,281 @@
...
+use crate::imperator::effect_validation::{EvB, EvBv, EvV};
+
+use Effect::*;
+
+pub fn scope_effect(name: &Token, _data: &Everything) -> Option<(Scopes, Effect)> {
+    let lwname = name.as_str().to_lowercase();
+
+    for (from, s, effect) in SCOPE_EFFECT {
+        if lwname == *s {
+            return Some((Scopes::from_bits_truncate(*from), *effect));
```
amtep, line 15: After replacing `u64` with `Scopes` below, this line will have just `*from` instead of `Scopes::from_bits_truncate(*from)`

src/imperator/tables/triggers.rs
```diff
@@ -0,0 +1,526 @@
...
+            && !data.item_exists(Item::PopType, s)
+        {
+            let msg = format!("could not find any {part}");
+            let info = "Possible valid options would be: num_of_$POPTYPE$ or num_of_$BUILDING$";
+            warn_info(name, ErrorKey::MissingItem, &msg, info);
+        }
+        return Some((Scopes::Country, Trigger::CompareValue)));
+    }
+    // This one is weird...the trigger is just Item::TechnologyTable with no suffix or prefix.
+    if data.verify_exists_implied(Item::TechnologyTable, "", name) {
```
amtep, line 40: This won't work because it makes every unknown trigger into a potential TechnologyTable trigger. It should just be
`if data.item_exists(Item::TechnologyTable, name.as_str())`

That way, it catches the TechnologyTable trigger if it is used, but lets other triggers pass unmolested.

src/imperator/tables/triggers.rs
```diff
@@ -0,0 +1,526 @@
...
+    }
+    std::option::Option::None
+}
+
+static TRIGGER_MAP: Lazy<FnvHashMap<String, (Scopes, Trigger)>> = Lazy::new(|| {
+    let mut hash = FnvHashMap::default();
+    for (from, s, trigger) in TRIGGER {
+        hash.insert(
+            (*s).to_string(),
+            (Scopes::from_bits_truncate(*from), Trigger::from_raw(trigger)),
```
amtep, line 51: This line will just be `(*from, *trigger)),` after making the changes below.

src/imperator/tables/triggers.rs
```diff
@@ -0,0 +1,526 @@
...
+        return Some((Scopes::Country, Trigger::CompareValue)));
+    }
+    // This one is weird...the trigger is just Item::TechnologyTable with no suffix or prefix.
+    if data.verify_exists_implied(Item::TechnologyTable, "", name) {
+        return Some((Scopes::Country, Trigger::CompareValue)));
+    }
+    std::option::Option::None
+}
+
+static TRIGGER_MAP: Lazy<FnvHashMap<String, (Scopes, Trigger)>> = Lazy::new(|| {
```
amtep, line 46: The String here can be `&'static str` to avoid cloning new Strings

src/imperator/tables/triggers.rs
```diff
@@ -0,0 +1,526 @@
...
+    let mut hash = FnvHashMap::default();
+    for (from, s, trigger) in TRIGGER {
+        hash.insert(
+            (*s).to_string(),
```
amtep, line 50: The .to_string() here can be removed when the change from String to `&'static str` is made above.

src/imperator/tables/triggers.rs
```diff
@@ -0,0 +1,526 @@
...
+            (Scopes::from_bits_truncate(*from), Trigger::from_raw(trigger)),
+        );
+    }
+    hash
+});
+
+/// LAST UPDATED VIC3 VERSION 1.3.6
+/// See `triggers.log` from the game data dumps
+/// A key ends with '(' if it is the version that takes a parenthesized argument in script.
+const TRIGGER: &[(u64, &str, RawTrigger)] = &[
```
amtep, line 60: This needs changes too: `u64` to `Scopes`, and `RawTrigger` to just `Trigger`, and every Scope value in the table should start with `Scopes::`

All to get rid of `RawTrigger` :)

Cargo.toml
```diff
@@ -12,7 +13,7 @@ name = "tiger-lib"
 version = "0.8.1"
 edition = "2021"
 license = "GPL-3.0-or-later"
-description = "Library used by the tools ck3-tiger and vic3-tiger. This library holds the bulk of the code for them. It can be built either for ck3-tiger with the feature ck3, or for vic3-tiger with the feature vic3, but not both at the same time."
+description = "Library used by the tools ck3-tiger, vic3-tiger, and imperator-tiger. This library holds the bulk of the code for them. It can be built either for ck3-tiger with the feature ck3, or for vic3-tiger with the feature vic3, or for imperator-tiger with the feature imperator, but not both at the same time."
```
amtep, line 16: "but not both" shoudl be "but not more than one" now

imperator-tiger.conf
```diff
@@ -0,0 +1,84 @@
...
+        # - text
+        # - ignore_keys_in_files
+    }
+
+}
+
+# Sometimes imperator-tiger makes a mistake in determining the scope types expected
+# by a scripted trigger or effect. Its conclusions can be overridden here.
+# Scope overrides can be ALL (to accept any scope) or a list separated by |,
+# like character|province|landed_title
```
amtep, line 50: This example could be different for imperator

src/imperator/data/ambitions.rs.bak
amtep, line 1: You could have just left these and removed their names from the `mod.rs` :)

src/lib.rs
```diff
@@ -21,11 +21,11 @@
 // validation of a field while handling the case of the field not existing.
 #![allow(clippy::blocks_in_if_conditions)]
 
-#[cfg(all(feature = "ck3", feature = "vic3"))]
-compile_error!("features \"ck3\" and \"vic3\" cannot be enabled at the same time");
+#[cfg(all(feature = "ck3", feature = "vic3", feature = "imperator"))]
```
amtep, line 24: We'll have to make this more complicated and check for any two or more feature flags :(

Unless I fix it so they can be on at the same time :)

src/validate.rs
```diff
@@ -751,6 +751,13 @@ pub fn validate_vic3_modifiers(vd: &mut Validator, sc: &mut ScopeContext) {
     });
 }
 
+#[cfg(feature = "imperator")]
+pub fn validate_imperator_modifiers(vd: &mut Validator, sc: &mut ScopeContext) {
```
amtep, line 755: This can probably be combined with validate_vic3_modifiers if they really both use the same modifier format

imperator-tiger/Cargo.toml
amtep, line 1: Needs more fields now that we're publishing on [crates.io]. `imperator-tiger` is not ready for that yet, but I'd like to do it ASAP to at least claim the name.

imperator-tiger/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "imperator-tiger"
+version = "0.6.1"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tiger-lib = { path = "..", features = ["imperator"] }
```
amtep, line 9: Should now have `version = "0.8.1",` and `default-features = false,` to turn off the default ck3 feature

amtep (approved):
Lots of things to change, but it can be merged as a Work In Progress

PR #17, opened 2023-08-10
Imperator validator changes

dementive:


src/imperator/data/governments.rs
```diff
@@ -0,0 +1,51 @@
...
+        vd.field_validated_block("base", |block, data| {
+            let vd = Validator::new(block, data);
+            validate_modifs(block, data, ModifKinds::Country, vd);
+        });
+        vd.field_validated_block("bonus", |block, data| {
+            let vd = Validator::new(block, data);
+            validate_modifs(block, data, ModifKinds::Country, vd);
+        });
+
+        vd.field_validated_block("color", validate_color);
```
amtep, line 49: Can it be a named color? Then you can do `vd.field_validated("color", validate_possibly_named_color);` instead.

src/imperator/data/levy_templates.rs
```diff
@@ -0,0 +1,35 @@
...
+        // TODO - How to do the units like that?
+        // levy_anatolian = {  #General Anatolian
+        //     default = no
+
+        //     light_infantry = 0.7
+        //     light_cavalry = 0.15
+
+        //     heavy_cavalry = 0.1
+        //     chariots = 0.05
+        // }
```
amtep, line 33: There's `vd.unknown_block_fields()`, which takes a closure with the key and the block, where you can then use data.verify_exists on the key and make a new Validator for the block.

Then probably do the same again inside that closure (using `vd.unknown_value_fields()` this time) to validate the keys and values for the unit types.

PR #18, opened 2023-08-21
Imperator - validator improvments

dementive:
- Added a new validator for imperator events

- Went through all the comments on the last PR and implemented all the suggestions that were made

- Filled in several missing localization fields from existing validators

### DIFF
--- a/src/imperator/data/deity.rs
+++ b/src/imperator/data/deity.rs
@@ -31,8 +31,13 @@ impl DbKind for Deity {
         let mut sc = ScopeContext::new(Scopes::Country, key);
 
         data.verify_exists(Item::Localization, key);
-        let loca = format!("omen_{key}_desc");
+        let key_str = key.to_string();
+        // Slice "deity_" off the front of the key
+        let diety_key = key_str.char_indices().nth(6).map_or(key_str.len(), |(i, _)| i);
+        let loca = format!("omen_{diety_key}");
+        let loca2 = format!("omen_{diety_key}_desc");
         data.verify_exists_implied(Item::Localization, &loca, key);
+        data.verify_exists_implied(Item::Localization, &loca2, key);
 
         vd.field_validated_block("trigger", |b, data| {
             validate_trigger(b, data, &mut sc, Tooltipped::No);
--- a/src/imperator/data/deity_categories.rs
+++ b/src/imperator/data/deity_categories.rs
@@ -23,7 +23,6 @@ impl DbKind for DeityCategory {
     fn validate(&self, key: &Token, block: &Block, data: &Everything) {
         let mut vd = Validator::new(block, data);
 
-        data.verify_exists(Item::Localization, key);
         let loca = format!("{key}").to_uppercase();
         data.verify_exists_implied(Item::Localization, &loca, key);
 
--- a/src/imperator/data/events.rs
+++ b/src/imperator/data/events.rs
@@ -1,0 +1,252 @@
+use std::path::PathBuf;
+use std::str::FromStr;
+
+use fnv::FnvHashMap;
+
+use crate::block::{Block, Field};
+use crate::context::{Reason, ScopeContext};
+use crate::desc::validate_desc;
+use crate::effect::{validate_effect, validate_effect_internal};
+use crate::everything::Everything;
+use crate::fileset::{FileEntry, FileHandler};
+use crate::helpers::dup_error;
+use crate::item::Item;
+use crate::lowercase::Lowercase;
+use crate::pdxfile::PdxFile;
+use crate::report::{error, error_info, warn_info, ErrorKey};
+use crate::scopes::Scopes;
+use crate::token::Token;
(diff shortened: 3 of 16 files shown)


src/imperator/data/deity.rs
```diff
@@ -31,8 +31,13 @@ impl DbKind for Deity {
         let mut sc = ScopeContext::new(Scopes::Country, key);
 
         data.verify_exists(Item::Localization, key);
-        let loca = format!("omen_{key}_desc");
+        let key_str = key.to_string();
+        // Slice "deity_" off the front of the key
+        let diety_key = key_str.char_indices().nth(6).map_or(key_str.len(), |(i, _)| i);
```
amtep, line 36: This is a job for strip_prefix :)

You can do it like
```
if let Some(no_deity_key) = key.as_str().strip_prefix("deity_") {
    ... check loca
} else {
    ... something -- maybe warn that key should start with deity_ ?
}
```

However it might be good to check if that's really what's happening here. I've never before seen Pdx manipulate key names like that. Could there be a separate omen_ key in a different folder, that just happens to have the same names as the deities? I've had a couple of cases like that in ck3.

Also check the spelling, it's deity not diety.

src/imperator/data/deity_categories.rs
```diff
@@ -23,7 +23,6 @@ impl DbKind for DeityCategory {
     fn validate(&self, key: &Token, block: &Block, data: &Everything) {
         let mut vd = Validator::new(block, data);
 
-        data.verify_exists(Item::Localization, key);
         let loca = format!("{key}").to_uppercase();
```
amtep, line 26: The `format` here is a no-op. You could do `let loca = key.as_str().to_uppercase()` instead.

src/imperator/data/events.rs
```diff
@@ -0,0 +1,252 @@
...
+];
+
+impl Event {
+    pub fn new(key: Token, block: Block) -> Self {
+        let mut expects_scope = Scopes::Country;
+        let mut expects_from_token = key.clone();
+
+        if let Some(event_type) = block.get_field_value("type") {
+            match event_type.as_str() {
+                "character_event" => {
```
amtep, line 145: Clippy will tell you to combine the arms with identical bodies, like `"character_event" | "minor_character_event" => `

Up to you whether to do that or add `#[allow(clippy::match_same_arms)]` here, it's mostly a style issue.

src/imperator/data/events.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        if !hidden {
+            vd.req_field("picture");
+        }
+        vd.field_item("picture", Item::EventPicture);
+
+        // TODO - there can be up to 3 left_portrait and 3 right_portrait fields in an event, not sure how to do this
+        vd.field_target("left_portrait", &mut sc, Scopes::Character);
+        vd.field_target("right_portrait", &mut sc, Scopes::Character);
```
amtep, line 207: You can do `self.block.count_keys("left_portrait")` to count them directly and then emit a warning. But what I tend to do instead is use `let mut count = 0`, then `vd.multi_field_validated_key_block`, and in its closure start with `count += 1` and then a check for `count == 4`. That way I can emit a warning exactly on the key of the 4th item, to say that there should be no more than 3.

Since here the logic for `left_portrait` and `right_portrait` is the same and the validator gets kinda long, it would make sense to start with `for field in &["left_portrait", "right_portrait"] {` and then repeat the validator only once.

src/imperator/data/levy_templates.rs
```diff
@@ -25,16 +25,12 @@ impl DbKind for LevyTemplate {
...
-        //     light_cavalry = 0.15
-
-        //     heavy_cavalry = 0.1
-        //     chariots = 0.05
-        // }
+        vd.unknown_block_fields(|key, block| {
+            vd.field_bool("default");
+            let mut vd = Validator::new(block, data);
+            vd.unknown_value_fields(|_, value| {
+                data.verify_exists(Item::UnitType, value);
```
amtep, line 32: If the entries look like `light_infantry = 0.7` then you should be doing `verify_exists` on the key, not the value, and then `value.expect_number()`

src/imperator/data/levy_templates.rs
```diff
@@ -25,16 +25,12 @@ impl DbKind for LevyTemplate {
...
-
-        //     light_infantry = 0.7
-        //     light_cavalry = 0.15
-
-        //     heavy_cavalry = 0.1
-        //     chariots = 0.05
-        // }
+        vd.unknown_block_fields(|key, block| {
+            vd.field_bool("default");
+            let mut vd = Validator::new(block, data);
```
amtep, line 30: There 2 lines should be switched

amtep (commented):
I found some things to fix, but since imperator-tiger doesn't work yet it's ok to merge this as-is.

amtep (approved):

PR #19, opened 2023-09-11
Finished up existing Imperator validators

dementive:
This should be all the validators for Imperator...except for the setup folder which will be done next before getting a working build for imperator going. There are some compiler errors with the `imperator-tiger` flag but they are all caused by unimplemented Item types and will be fixed when everything get added to everything.rs, ck3 and vic3 still compile fine.


src/imperator/data/laws.rs
```diff
@@ -1 +1,94 @@
...
+// TODO - Check if this actually needs to be submitted, same for the Culture Item
+inventory::submit! {
+    ItemLoader::Normal(GameFlags::Imperator, Item::Law, Law::add)
+}
+
+impl Law {
+    pub fn add(db: &mut Db, key: Token, block: Block) {
+        db.add(Item::Law, key, block, Box::new(Self {}));
+    }
+}
```
amtep, line 59: Yeah these shouldn't be here

src/imperator/data/levy_templates.rs
```diff
@@ -26,10 +24,12 @@ impl DbKind for LevyTemplate {
         let mut vd = Validator::new(block, data);
 
         vd.unknown_block_fields(|key, block| {
-            vd.field_bool("default");
             let mut vd = Validator::new(block, data);
+
+            vd.field_bool("default");
             vd.unknown_value_fields(|_, value| {
```
amtep, line 30: You should bind `key` here, since you check it below

src/imperator/data/military_traditions.rs
```diff
@@ -1 +1,105 @@
...
+
+inventory::submit! {
+    ItemLoader::Normal(GameFlags::Imperator, Item::MilitaryTradition, MilitaryTradition::add)
+}
+
+impl MilitaryTradition {
+    pub fn add(db: &mut Db, key: Token, block: Block) {
+        db.add(Item::MilitaryTradition, key, block, Box::new(Self {}));
+    }
+}
```
amtep, line 69: These shouldn't be here, since MilitaryTradition is added by MilitaryTraditionTree::add

src/imperator/data/missions.rs
```diff
@@ -1 +1,146 @@
...
+
+inventory::submit! {
+    ItemLoader::Normal(GameFlags::Imperator, Item::MissionTask, MissionTask::add)
+}
+
+impl MissionTask {
+    pub fn add(db: &mut Db, key: Token, block: Block) {
+        db.add(Item::MissionTask, key, block, Box::new(Self {}));
+    }
+}
```
amtep, line 89: Same story here :)

src/imperator/data/missions.rs
```diff
@@ -1 +1,146 @@
...
+
+        vd.field_validated_block("requires", |b, data| {
+            data.verify_exists(Item::MissionTask, key);
+        });
+
+        vd.field_validated_block("prevented_by", |b, data| {
+            data.verify_exists(Item::MissionTask, key);
+        });
+
+        // TODO - there is a better way to do this, scope:province just has to be present inside of the highlight block AND inside scope:province context is province
```
amtep, line 113: I think you can just use `sc.define_name` to add "province" as a Province scope, and then validate "highlight" as a normal trigger.

I don't think the game actually requires that you start with `scope:province = {` in the trigger? It's just the logical thing to do.

amtep (commented):
I left some comments :) I didn't merge it because you're still working on it.

src/imperator/data/combat_tactics.rs
```diff
@@ -0,0 +1,72 @@
...
+        vd.field_item("sound", Item::Sound);
+
+        vd.unknown_value_fields(|key, value| {
+            data.verify_exists(Item::CombatTactic, key);
+            value.expect_number();
+        });
+
+        vd.field_validated_block("effective_composition", |block, data| {
+            data.verify_exists(Item::Unit, key);
+            //value.expect_number();
```
amtep, line 69: Does this need to be verified?

src/imperator/data/inventions.rs
```diff
@@ -1 +1,84 @@
...
+        let mut sc = ScopeContext::new(Scopes::Country, key);
+
+        data.verify_exists(Item::Localization, key);
+        let loca = format!("{key}_desc");
+        data.verify_exists_implied(Item::Localization, &loca, key);
+
+        vd.field_bool("keystone");
+        vd.field_choice("icon_override", &["gw_icon", "war"]);
+
+        // TODO - check if this is the correct way to validate this:
```
amtep, line 60: Yeah, that's right. Assuming `agressive_expansion_impact_inv_5` is an invention

src/imperator/data/setup_main.rs
```diff
@@ -0,0 +1,295 @@
...
+
+fn validate_treasures(block: &Block, data: &Everything) {
+    let mut vd = Validator::new(block, data);
+    vd.field_validated_block("database", |block, data| {
+        let mut vd = Validator::new(block, data);
+        for (_, block) in vd.integer_blocks() {
+            let mut vd = Validator::new(block, data);
+            vd.field_item("key", Item::Localization);
+            vd.choice(DLC_IMPERATOR);
+            vd.field("icon"); // TODO - icon can be any icon declared in "gfx/interface/icons/treasures", how to check that?
```
amtep, line 86: You can do a verify_exists_implied on Item::File to check if an icon's file exists

src/imperator/data/setup_main.rs
```diff
@@ -0,0 +1,295 @@
...
+            validate_trade(block, data);
+        });
+        vd.field_validated_block("great_work_manager", |block, data| {
+            validate_great_works(block, data);
+        });
+    }
+}
+
+impl FileHandler<Block> for SetupMain {
+    // TODO - Is this impl needed? Copied it from src/vic3/data/history.rs
```
amtep, line 58: This depends om whether SetupMain follows the key = { ... } format of most game item types. If it does, you can use the DbKind validate approach. If it's too different, you have to use the FileHandler approach which is more manual (and is an older API)

src/imperator/data/setup_main.rs
```diff
@@ -0,0 +1,295 @@
...
+            let mut vd = Validator::new(block, data);
+            vd.field_item("first", Item::Country);
+            vd.field_item("second", Item::Country);
+            if field == "dependency" {
+                vd.field_item("subject_type", Item::SubjectType);
+            }
+        });
+    }
+    /* 
+    TODO - imperator - This block also has a trade_access section with irregular syntax, not sure how to do this, it has two nested Item::Country basically
```
amtep, line 125: You could use the unknown_block_field approach (honestly I don't remember if that's the right method name) and check that the key is a country, then create an inner validator and do the same to that one .

src/imperator/data/setup_main.rs
```diff
@@ -0,0 +1,295 @@
...
+fn validate_roads(block: &Block, data: &Everything) {
+    // This is just 2 provinces connected like
+    /*
+        1 = 2
+        100 = 110
+    */
+    let mut vd = Validator::new(block, data);
+
+    vd.unknown_value_fields(|key, value| {
+        value.expect_number();
```
amtep, line 172: Might be nice to check that the province ids are defined

src/imperator/data/setup_main.rs
```diff
@@ -0,0 +1,295 @@
...
+
+            great_work_effect_selections={
+                {
+                    great_work_effect="gw_effect_omen_doctrine"
+                    great_work_effect_tier = "gw_effect_tier_4"
+                }
+            }
+        }
+
+        How to match the loose brackets in the great_work_components and great_work_effect_selections blocks???
```
amtep, line 287: Those are known as "loose blocks" and there are some methods for iterating over them. Might be as simple as validate_blocks, but my memory is spotty.

src/item.rs
```diff
@@ -80,6 +80,7 @@ pub enum Item {
     GuiLayer,
     GuiTemplate,
     GuiType,
+    LawGroup,
```
amtep, line 83: Please also move it in the mapping of Item to folders below :)

amtep (commented):
I left some comments :) sorry it took so long

src/imperator/data/combat_tactics.rs
dementive: The script appears like this:
```
    effective_composition = {
        <unit_type> = 0
        <unit_type> = 0
        <unit_type> = 0
        <unit_type> = 1.0
        <unit_type> = 1.0
        <unit_type> = 0
        <unit_type> = 0.0
        <unit_type> = 0.75
        <unit_type> = 2.0
    }
```
So I think this is probably the way to do it, although there may be something better

src/imperator/data/setup_main.rs
dementive: I think the DbKind validate approach will be enough here in that case, will remove the FileHandler impl then

src/imperator/data/combat_tactics.rs
amtep: Yeah I was wondering why `value.expect_number()` is commented out :)

amtep:
I rebased this and pushed it to main

PR #23, opened 2024-01-17
Complex comparison triggers

my4ng:
This PR adds support for special comparison triggers, i.e. "comparison trigger that needs a parsable string parameter that will return its value in the context it is used", in the `"trigger(arg)"` syntax. 

For example, `"cultural_acceptance(scope:target_culture)" > 50`.


src/vic3/tables/triggers.rs
```diff
@@ -805,3 +810,24 @@ const TRIGGER: &[(Scopes, &str, Trigger)] = &[
...
+        }
+        hash
+    });
+
+/// LAST UPDATED VIC3 VERSION N/A
+/// See `triggers.log` from the game data dumps
+/// `(inscopes, trigger name, argtype)`
+/// Currently only works with single argument triggers
+// TODO Verify triggers
+const TRIGGER_SPECIAL_VALUE: &[(Scopes, &str, Trigger)] = &[];
```
amtep, line 833: Hmm this is a regression since you did remove them from handle_argument. Can you put those in the table?

amtep (changes requested):

src/vic3/tables/triggers.rs
my4ng: Some of the argument validation requires updating vic3 as a whole to 1.5, i.e. adding new items and scopes. I will leave out those as `UncheckedValue` for now and leave a TODO for when it is updated.

my4ng: In addition, I think entries from the `event_targets.log` which have output scope of value (such as `num_mutual_trade_route_levels_with_country` should *not* be mixed with the entries from `triggers`, but within their own table in `scopes.rs` instead.

my4ng:
There are still a number of unresolved issues with this PR which is *not* ready for merge.

- [ ] Scope to scope with value output type (some were erroneously placed previously)
- [ ] Values stored in variables

amtep (dismissed):

src/trigger.rs
```diff
@@ -969,114 +974,101 @@ pub fn validate_target_ok_this(
...
-            err(ErrorKey::UnknownField).msg(msg).opt_info(opt_info).loc(part).push();
-            sc.close();
-            return;
         }
     }
     let (final_scopes, because) = sc.scopes_reason();
     if !outscopes.intersects(final_scopes | Scopes::None) {
         let part = &part_vec[part_vec.len() - 1];
         let msg = format!("`{part}` produces {final_scopes} but expected {outscopes}");
-        if part == because.token() && part.loc == because.token().loc {
```
amtep, line 1074: I remember putting this in because otherwise it led to spurious extra message blocks that pointed to the same token as the original block

src/trigger.rs
```diff
@@ -1091,65 +1083,298 @@ pub fn validate_target(token: &Token, data: &Everything, sc: &mut ScopeContext,
...
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use crate::{emit_reports, FileKind, Loc};
+
+    use super::*;
+    #[test]
+    fn test_partition() {
+        let path = "./test.txt";
```
amtep, line 1363: This is better placed in tests/files/ with a descriptive name

src/trigger.rs
```diff
@@ -1091,65 +1083,298 @@ pub fn validate_target(token: &Token, data: &Everything, sc: &mut ScopeContext,
...
+            PathBuf::from_str(path).unwrap(),
+            FileKind::Mod,
+            PathBuf::from_str(path).unwrap(),
+        );
+        loc.line += 1;
+        loc.column += 1;
+        let str = std::fs::read_to_string(path).unwrap();
+        let token = Token::new(&str, loc);
+        let partitioned = partition(&token);
+        dbg!(token);
```
amtep, line 1374: Debug output is swallowed by the test framework if the test succeeds. Common practice is to use assert macros to assert that the vector and it's tokens are what's expected.

src/item.rs
```diff
@@ -934,6 +935,8 @@ impl Item {
             #[cfg(feature = "vic3")]
             Item::DynamicCountryName => "common/dynamic_country_names/",
             #[cfg(feature = "vic3")]
+            Item::EventCategory => "events/",
```
amtep, line 938: Since it's defined in an internal table  the path here should be "". The path should be events iff the item values are collected from data in events/

amtep (changes requested):
Looks good :) Just some small notes. I also haven't tried it yet.

Housekeeping note: I prefer to resolve conflicts by rebasing the feature branch, not by making changes in a merge commit.

src/trigger.rs
my4ng: I am more or less using this as a test playground to see if the output (including the error reports) meet my expectation, so kind of difficult to use `assert` here

src/trigger.rs
my4ng: It's going to be difficult as a lot of the functions and struct types are private

amtep: I meant the file not the code :)

ck3-tiger/src/bin/scan-mod-ck3-tiger.rs
```diff
@@ -47,7 +47,7 @@ fn main() -> Result<()> {
     everything.load_all();
 
     let mut grand_json = json!({});
-    for itype in Item::iter() {
+    for itype in <Item as IntoEnumIterator>::iter() {
```
amtep, line 50: What does this do?

src/validate.rs
```diff
@@ -15,7 +15,7 @@ use crate::game::Game;
 use crate::item::Item;
 use crate::lowercase::Lowercase;
 use crate::report::{err, error, fatal, old_warn, report, warn, Confidence, ErrorKey, Severity};
-use crate::scopes::{scope_prefix, scope_to_scope, Scopes};
+use crate::scopes::{scope_prefix, scope_to_scope};
```
amtep, line 18: Why this change? I prefer to import everything at the top so that there are no surprises

my4ng: It is unused import when building vic3-tiger

amtep: I see :)

amtep (approved):

PR #25, opened 2024-01-30
CK3 Fix for 1.11.3

my4ng:
A few bugfixes for CK3 1.11.3 (peacock)


amtep (changes requested):
Hmm instead of having a revert in the commit history, I'd rather see a rebase where that commit is deleted

amtep (approved):

PR #26, opened 2024-02-01
feat: add culture parameters and layers validation

amtep:


src/fileset.rs
```diff
@@ -449,17 +453,56 @@ impl Fileset {
...
+                .msg(msg)
+                .loc(token)
+                .push();
+        }
+    }
+
     #[cfg(feature = "ck3")] // vic3 happens not to use
     pub fn verify_exists(&self, file: &Token) {
-        self.mark_used(&file.as_str().replace("//", "/"));
+        self.mark_used(file.as_str());
```
amtep, line 497: Why this change? The stripping of `//` was done because paths containing that occur in the wild.

src/fileset.rs
```diff
@@ -449,17 +453,56 @@ impl Fileset {
...
         if !self.exists(file.as_str()) {
             let msg = "referenced file does not exist";
             report(ErrorKey::MissingFile, Item::File.severity()).msg(msg).loc(file).push();
         }
     }
 
     pub fn verify_exists_implied(&self, file: &str, t: &Token, max_sev: Severity) {
-        self.mark_used(&file.replace("//", "/"));
+        self.mark_used(file);
```
amtep, line 505: Same

src/fileset.rs
```diff
@@ -470,7 +513,7 @@ impl Fileset {
     }
 
     pub fn verify_exists_implied_crashes(&self, file: &str, t: &Token) {
-        self.mark_used(&file.replace("//", "/"));
+        self.mark_used(file);
```
amtep, line 516: Same

src/ck3/data/cultures.rs
```diff
@@ -252,9 +253,47 @@ impl DbKind for CultureTradition {
...
+                for path in block.iter_values_warn() {
+                    layer_path.push(path.as_str());
+                }
+            }
+
+            let mut vd = Validator::new(block, data);
+            vd.unknown_value_fields(|key, value| {
+                if let Some(layer_idx) = key.expect_integer().and_then(|i| {
+                    // short circuit evaluation so no panic
+                    if i >= 0 || (i as usize) < layer_path.len() {
```
amtep, line 283: I think this needs to be `&&` not `||`

src/ck3/data/characters.rs
```diff
@@ -301,125 +301,270 @@ impl Character {
...
+                    vd.req_field("death_reason");
+                    vd.field_item("death_reason", Item::DeathReason);
+                    vd.field_item("killer", Item::Character);
+                    Some((Death, key))
+                }
+                "effect" => {
+                    validate_effect(block, data, sc, Tooltipped::No);
+                    None
+                }
+                _ => {
```
amtep, line 444: Unfortunately you can't replace `validate_effect_internal` with this. Ck3 allows arbitrary effects directly in the history, without an `effect` block.

This also means you can't skip the use of a Validator that marks the non-effect fields as "known" before calling `validate_effect_internal`

amtep (commented):
Some structural problems unfortunately.

src/ck3/data/characters.rs
```diff
@@ -301,125 +301,270 @@ impl Character {
...
-            if !token.as_str().starts_with("c_") {
-                error(token, ErrorKey::Validation, "capital must be a county");
+    fn validate_life(
+        character: &Token,
+        mut traits: FnvHashSet<Token>,
+        life_events: Vec<LifeEvent>,
+    ) {
+        let mut birth = None;
+        let mut death = None;
+        let mut spouse = None;
```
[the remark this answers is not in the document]
my4ng: Yes you are right. I do remember the num_of_spouses in the culture previously. I will make it a list then

src/ck3/data/characters.rs
```diff
@@ -301,125 +301,270 @@ impl Character {
...
+        let mut claims = FnvHashSet::<Token>::default();
+        let mut employed = false;
+
+        for LifeEvent { date, index: _, token, event } in life_events.into_iter() {
+            use LifeEventType::*;
+
+            if birth.is_none() && event != Birth {
+                let msg = format!("{character} is not born yet");
+                let mut loc = token.loc.clone();
+                loc.column = 0;
```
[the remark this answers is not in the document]
my4ng: Otherwise only the token rather than the line  is highlighted, which can be confusing, even though it is the whole action that is invalid. For example, add_pressed_claim would highlight the title, which could be misleading

src/ck3/data/characters.rs
my4ng: Okay. I didn’t see any loose effect in the vanilla files. Just for reference, this function is validating each line individually rather than as a block though. So the highlighted line is actually evaluating the effect block.

amtep: My comment was attached to the `_ => {` line :) Which starts the rejection of all unknown effects. It's a pity that github only shows the lines above and not below

src/ck3/data/characters.rs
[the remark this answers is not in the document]
amtep: Tricky :) Like, `remove_claim` should have the claim value highlighted if the problem is that the claim was never added, and it should have the `remove_claim` key highlighted if the problem was that it's happening after death.

I'm still thinking this over. I don't really like the departure from tiger's usual habit of highlighting the tokens. `column = 0` is supposed to be a last resort if no token is available.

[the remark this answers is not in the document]
my4ng: Yes I have corrected it to use the token for non-birth/death warnings now. They should now point to the relevant claim/trait/etc.

src/ck3/data/characters.rs
```diff
@@ -435,50 +413,54 @@ impl Character {
...
+                _ => (),
             },
         };
 
-        result.map(|(etype, token)| (etype, token.clone()))
+        // unknown effect field
+        validate_effect_field(
+            &crate::lowercase::Lowercase::empty(),
+            key,
+            crate::block::Comparator::Equals(crate::block::Eq::Single),
```
amtep, line 430: Importing just `Eq::Single` is done in some other files too, you can see how it's usually done

src/ck3/data/characters.rs
```diff
@@ -435,50 +413,54 @@ impl Character {
...
-                warn(ErrorKey::Validation).msg(msg).loc(loc).push();
+                warn(ErrorKey::History).msg(msg).loc(loc).push();
             }
 
             if let Some((date, death_loc)) = &death {
-                let msg = format!("{character} has died already on {date}");
-                let mut loc = token.loc.clone();
-                loc.column = 0;
-                warn(ErrorKey::Validation).msg(msg).loc(loc).loc(death_loc, "from here").push();
+                // Pothumous
```
amtep, line 457: Typo :)
Though I'd say the whole comment is not necessary

src/ck3/data/characters.rs
```diff
@@ -487,84 +469,58 @@ impl Character {
...
                     }
-                    spouse = None;
                 }
                 Employed => employed = true,
+                Unemployed => employed = false,
                 GiveCouncilPosition => {
                     if !employed {
-                        let msg = format!("{character} is not currently employed");
-                        warn(ErrorKey::Validation).msg(msg).loc(token).push();
+                        let msg = format!("{character} was not employed");
```
amtep, line 502: I think this message might be confusing because people will think having a council position means being employed. This is really about whether the character is at someone's court, right? A better message might be "{character} not at a court".

Btw many of these messages might be improved with an "at {date}", giving the date of the current history block. It more clearly tells what's the problem.

src/effect.rs
```diff
@@ -100,289 +100,295 @@ pub fn validate_effect_internal<'a>(
...
-        if let Some(effect) = data.get_effect(key) {
-            match bv {
-                BV::Value(token) => {
-                    if !effect.macro_parms().is_empty() {
-                        fatal(ErrorKey::Macro).msg("expected macro arguments").loc(token).push();
-                    } else if !token.is("yes") {
-                        old_warn(token, ErrorKey::Validation, "expected just effect = yes");
-                    }
-                    effect.validate_call(key, data, sc, tooltipped);
+        validate_effect_field(caller, key, cmp, bv, data, sc, tooltipped);
```
amtep, line 103: This is a nice solution :)

amtep (commented):
Looks good to me now

The first commit (culture layers) still has some comments though

src/ck3/tables/modifs.rs
```diff
@@ -29,12 +40,10 @@ pub fn lookup_modif(name: &Token, data: &Everything, warn: Option<Severity>) ->
...
         "_ai_vengefulness",
         "_ai_zeal",
-        "_different_culture_opinion",
-        "_different_faith_opinion",
-        "_same_culture_opinion",
-        "_same_faith_opinion",
     ] {
         if let Some(s) = name.as_str().strip_suffix(sfx) {
+            // Verify modifier format
+            data.verify_exists(Item::ModifierFormat, name);
```
amtep, line 46: I checked my memory and validate_modif already does this

https://github.com/amtep/ck3-tiger/blob/d93f8da63053dbdad126f95f825e3b812f637cd8/src/modif.rs#L105

amtep:
The individual fixes have been cherry-picked and pushed

PR #27, opened 2024-02-06
Macro Map

my4ng:
- Use `Option<NonZeroU32>` instead of `Option<Arc<Loc>>` for `Loc` link field
- Make `Loc` `Copy`

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -411,10 +411,9 @@ impl Block {
     }
 
     /// Expand a block that has macro parameters by substituting arguments for those parameters,
-    /// then re-parsing the script.
-    pub fn expand_macro(&self, args: &[(&str, Token)], link: &Token) -> Option<Block> {
-        let link_index = MACRO_MAP.get_index(link.loc);
-
+    /// then re-parsing the script, that links the expanded content back to `loc`.
+    pub fn expand_macro(&self, args: &[(&str, Token)], loc: Loc) -> Option<Block> {
+        let link_index = MACRO_MAP.get_or_insert_loc(loc);
         if let Some(source_box) = &self.source {
             let (source, local_macros) = source_box.as_ref();
             let mut content = Vec::new();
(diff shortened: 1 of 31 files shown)


src/data/scripted_effects.rs
```diff
@@ -37,6 +38,9 @@ impl Effects {
             err(ErrorKey::NameConflict).strong().msg(msg).loc(key).push();
         } else {
             let scope_override = self.scope_overrides.get(key.as_str()).copied();
+            if block.source.is_some() {
```
amtep, line 41: This pre-seeding might not be useful and it makes the API of expand_macro more complicated. It made sense when you could say the macro-map was read only during validation, but that's not the case now.

amtep (commented):
Just one observation :)

I haven't tried running it yet

src/data/scripted_effects.rs
my4ng: Although this is not strictly necessary, it is good to load all the scripted things here rather than later during the validation stage, in order to lower lock contention, especially since it requires a write lock.

src/block.rs
```diff
@@ -414,25 +413,26 @@ impl Block {
     /// Expand a block that has macro parameters by substituting arguments for those parameters,
     /// then re-parsing the script.
     pub fn expand_macro(&self, args: &[(&str, Token)], link: &Token) -> Option<Block> {
-        let link = Arc::new(link.loc.clone());
+        let link_index = MACRO_MAP.get_index(link.loc);
```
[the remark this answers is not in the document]
my4ng: `expand_macro` is only called by scripted things such as scripted_effects, for which the keys have been added already as in the comment below. If a new type of scripted item is added, then it is required that its key being added to the macro map already, assuming it calls expand_macro with its key as link too.

In fact, considering that's the only place link is used, it is probably better to require the caller to pass `link_index: NonZeroU32` to ensure that it has a valid link loc.

amtep:
I squashed some of the commits but didn't make changes

PR #31, opened 2024-02-12
Vic3 modifs

amtep:
vic3: fix the modifier: prefix and accept user-defined modifs

### DIFF
--- a/src/ck3/scopes.rs
+++ b/src/ck3/scopes.rs
@@ -7,8 +7,7 @@ use once_cell::sync::Lazy;
 
 use crate::everything::Everything;
 use crate::helpers::display_choices;
-use crate::scopes::Scopes;
-use crate::trigger::Trigger;
+use crate::scopes::{ArgumentValue, Scopes};
 
 // LAST UPDATED CK3 VERSION 1.11.3
 pub fn scope_from_snake_case(s: &str) -> Option<Scopes> {
@@ -468,11 +467,11 @@ const SCOPE_TO_SCOPE: &[(Scopes, &str, Scopes)] = &[
 ];
 
 #[inline]
-pub fn scope_prefix(name: &str) -> Option<(Scopes, Scopes, Trigger)> {
+pub fn scope_prefix(name: &str) -> Option<(Scopes, Scopes, ArgumentValue)> {
     SCOPE_PREFIX_MAP.get(name).copied()
 }
 
-static SCOPE_PREFIX_MAP: Lazy<FnvHashMap<&'static str, (Scopes, Scopes, Trigger)>> =
+static SCOPE_PREFIX_MAP: Lazy<FnvHashMap<&'static str, (Scopes, Scopes, ArgumentValue)>> =
     Lazy::new(|| {
         let mut hash = FnvHashMap::default();
         for (from, s, to, argument) in SCOPE_PREFIX.iter().copied() {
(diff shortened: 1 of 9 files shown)


dragon-archer:
I think it may be better if we could read from the modifier type database `common/modifier_types` rather than use localization, as I've tested this before that the engine won't accept unknown modifier types, and actually it's normal to forget adding the localization.

dragon-archer:
Also, may be it's possible to read all modifier types directly from the database? Thus the validator can be easily adapted to newer versions of the game without needing to add those modifer types manually, though it may hurt the performance to some degree

amtep:
> Also, may be it's possible to read all modifier types directly from the database? Thus the validator can be easily adapted to newer versions of the game without needing to add those modifer types manually, though it may hurt the performance to some degree

The database dump contains an explicit list of modifs, but the actual modifs present in the game depends on some things the mod can change, such as which goods exist and which combat unit types exist. So the modif list from the database dump is not necessarily accurate or complete.

The table in tiger's source code is just the leftovers after all the patterned modifs have been handled

dragon-archer:
> > Also, may be it's possible to read all modifier types directly from the database? Thus the validator can be easily adapted to newer versions of the game without needing to add those modifer types manually, though it may hurt the performance to some degree
> 
> The database dump contains an explicit list of modifs, but the actual modifs present in the game depends on some things the mod can change, such as which goods exist and which combat unit types exist. So the modif list from the database dump is not necessarily accurate or complete.
> 
> The table in tiger's source code is just the leftovers after all the patterned modifs have been handled

Perhaps I should make it clearer that there're actually 2 databases.
One is dumped from the console comand `script_docs`, which contains a list of all ***currently*** accepted modifier types, so if you run the command with a clear instance it outputs all *builtin* types, however if you run the command with mod enabled that add more goods, then the dumped list is also expanded.
Another is entries listed in `common/modifier_types`, and this is the actually list of all modifier types, the engine also read from this database to generate the first one.

So, there're no `builtin` modifier types actually, all modifier types are scriptable (except a few *mechanic* ones like `limited_to_frontier_colonization`). It's true that most modifer types are auto generated from goods or building types, but to make them function properly you still have to add them to `modifer_types` and specify their attributes (good or bad, integer or fraction)

dragon-archer:
In case of usage, I personally think that the first database can be used for *static check*, while the second is better for *dynamic check*

For example, in my project [paradox-highlight](https://github.com/dragon-archer/paradox-highlight), I choose the first one for highlighting since I can't scan the second database (It's based on static regex)

However, vic3-tiger is able to scan the second database, so I think it's better to choose the second one for better accuracy

amtep:
> It's true that most modifer types are auto generated from goods or building types, but to make them function properly you still have to add them to modifer_types and specify their attributes (good or bad, integer or fraction) 

However, I have found quite a few modifs that are used in vanilla but don't have a modifier_types definition. Example: `country_officers_voting_power_add`

I'm not sure whether that's an error or whether it works ok.

So I'd rather have one validation for whether something is a valid modif, and then separate validations to check if it has a modifier_type definition and whether it has the appropriate loca entries. The check whether something is a valid modif can also be specific about what's wrong, for example if you write `country_oficers_voting_power_add` it can say that pop type `oficers` does not exist, rather than a generic message.

But since user-defined modifs are allowed, I added a heuristic to accept those. First it was whether the `modifier_` loca exists, but after your comments I changed it to check whether a modifier_types entry exists.

dragon-archer:
> However, I have found quite a few modifs that are used in vanilla but don't have a modifier_types definition. Example: `country_officers_voting_power_add`

Yes you're right, I've tested it and it does work. I have to say the engine is so miraculous

> So I'd rather have one validation for whether something is a valid modif, and then separate validations to check if it has a modifier_type definition and whether it has the appropriate loca entries. The check whether something is a valid modif can also be specific about what's wrong, for example if you write `country_oficers_voting_power_add` it can say that pop type `oficers` does not exist, rather than a generic message.

This is a good idea.

> But since user-defined modifs are allowed, I added a heuristic to accept those. First it was whether the `modifier_` loca exists, but after your comments I changed it to check whether a modifier_types entry exists.

Well, I agree this might be the best solution.

my4ng (approved):

PR #36, opened 2024-02-16
ck3: validate history wars

amtep:


src/ck3/data/wars.rs
```diff
@@ -0,0 +1,94 @@
...
+        vd.field_list_items("targeted_titles", Item::Title);
+        vd.field_item("casus_belli", Item::CasusBelli);
+        vd.field_list_items("attackers", Item::Character);
+        vd.field_list_items("defenders", Item::Character);
+        vd.field_item("claimant", Item::Character);
+
+        vd.unknown_block_fields(|key, block| {
+            if data.item_exists(Item::Character, key.as_str()) {
+                let mut vd = Validator::new(block, data);
+                vd.validate_history_blocks(|_date, block, data| {
```
my4ng, line 83: can also validate if the date is within the start and end date

src/ck3/data/wars.rs
```diff
@@ -0,0 +1,94 @@
...
+impl War {
+    pub fn new(block: Block) -> Self {
+        Self { block }
+    }
+
+    fn validate(&self, data: &Everything) {
+        let mut vd = Validator::new(&self.block, data);
+
+        vd.req_field("start_date");
+        vd.req_field("end_date");
```
my4ng, line 68: Other fields (`targeted_titles`, `casus_belli`, `attackers`, `defenders`) would likely be required as well. Best to verify

my4ng (changes requested):
looks good just the two minor things

src/ck3/data/wars.rs
my4ng: same applies for start date should be before end date

my4ng (approved):

PR #37, opened 2024-02-16
Validate conflicting doctrines

amtep:


src/ck3/data/religions.rs
```diff
@@ -69,7 +71,44 @@ impl DbKind for Religion {
...
+                                format!("{doctrine} and {other_doctrine} are both from {category}");
+                            err(ErrorKey::Conflict)
+                                .msg(msg)
+                                .loc_msg(doctrine, "doctrine")
+                                .loc(other_doctrine, "earlier doctrine")
+                                .push();
+                        }
+                    } else if picks == (seen.len() as i64) {
+                        let msg =
+                            format!("religion has more than {picks} doctrines from {category}");
```
my4ng, line 103: perhaps we can link it to the `number_of_picks = X` line by storing a loc alongside the actual value

src/ck3/data/religions.rs
```diff
@@ -69,7 +71,44 @@ impl DbKind for Religion {
...
+                        if doctrine == other_doctrine {
+                            let msg = format!("duplicate doctrine {doctrine}");
+                            warn(ErrorKey::DuplicateField)
+                                .msg(msg)
+                                .loc_msg(doctrine, "doctrine")
+                                .loc(other_doctrine, "earlier doctrine")
+                                .push();
+                        } else {
+                            let msg =
+                                format!("{doctrine} and {other_doctrine} are both from {category}");
```
my4ng, line 94: this can be a bit ambiguous. best to mention that you can only pick one doctrine from this category

src/ck3/data/religions.rs
```diff
@@ -69,7 +71,44 @@ impl DbKind for Religion {
...
+            if let Some(category) = data.doctrines.category(vd.value().as_str()) {
+                let doctrine = vd.value();
+                if let Some(seen) = categories.get_mut(category.as_str()) {
+                    let picks = data.doctrines.number_of_picks(category.as_str());
+                    #[allow(clippy::cast_possible_wrap)]
+                    if picks == 1 {
+                        // SAFETY: we never push empty vecs into this hash
+                        let other_doctrine = &seen[0];
+                        if doctrine == other_doctrine {
+                            let msg = format!("duplicate doctrine {doctrine}");
```
my4ng, line 86: I don't think duplicate doctrine is being checked for when picks > 1?

my4ng (changes requested):
some minor issues

src/ck3/data/religions.rs
amtep: Hmm I didn't do that because it's the default. There's basically only one category that gets number_of_picks and that's the core tenets. So picks > 1 is the exception

amtep: Oh, I could put that in the info though

src/ck3/data/religions.rs
amtep: Yeah it's worth checking it there too

src/ck3/data/religions.rs
amtep: Yeah though I think I'll do it so that picks is simply preserved as a Token and then decoded where it's used.
That's the intended use of Token, at leas

src/ck3/data/religions.rs
```diff
@@ -168,6 +123,60 @@ impl DbKind for Religion {
...
+                        .push();
+                } else if picks == 1 {
+                    // SAFETY: we never push empty vecs into this hash
+                    let other_doctrine = &seen[0];
+                    let msg = format!("{doctrine} and {other_doctrine} are both from {category}");
+                    let info = format!("{category} only allows 1 pick");
+                    err(ErrorKey::Conflict)
+                        .msg(msg)
+                        .info(info)
+                        .loc_msg(doctrine, "doctrine")
```
my4ng, line 152: I would have considered using `opt_loc` but this is fine too

my4ng (approved):

PR #39, opened 2024-02-19
Add argument for custom config file

dementive:
Adds a new --config argument to each program to allow the user to load a custom configuration file. I tested with all 3 versions and it seems to be working for all of them. You may want to implement validate_config_file() in a slightly different way but it looks like it works fine to me.


amtep (approved):
Looks good to me

PR #40, opened 2024-02-19
ck3: validate courtier_guest_management

amtep:


src/ck3/data/courtier_guest_mgmt.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Clone, Debug)]
+pub struct CourtierGuestManagement {}
+
+inventory::submit! {
+    ItemLoader::Normal(GameFlags::Ck3, Item::CourtierGuestManagement, CourtierGuestManagement::add)
+}
+
+impl CourtierGuestManagement {
+    pub fn add(db: &mut Db, key: Token, block: Block) {
+        db.add(Item::NameList, key, block, Box::new(Self {}));
```
my4ng, line 25: Shouldn't this be `Item::CourtierGuestManagement`?

src/ck3/data/courtier_guest_mgmt.rs
```diff
@@ -0,0 +1,88 @@
...
+                    });
+                    vd.field_validated_key("desc", |key, bv, data| {
+                        validate_desc(bv, data, &mut sc_builder(key));
+                    });
+                });
+            });
+        } else {
+            let msg = "expected either `courtier_management` or `guest_management`";
+            err(ErrorKey::Validation).msg(msg).loc(key).push();
+            return;
```
my4ng, line 85: don't see why we need the `return` there

my4ng (changes requested):

src/ck3/data/courtier_guest_mgmt.rs
amtep: Oops yes

src/ck3/data/courtier_guest_mgmt.rs
amtep: Ah yeah that was left over after restructuring the code

PR #41, opened 2024-02-19
Fix several false positives for imperator

dementive:
There were lots of scopes and modifiers types that the logs were incorrect on and several validators that I made mistakes on the first pass. There are still lots of false positives but I tried to fix most of the obvious ones here.


src/imperator/data/ambitions.rs
```diff
@@ -36,7 +36,7 @@ impl DbKind for Ambition {
 
         vd.req_field("can_be_picked");
         vd.req_field("finished_when");
-        vd.req_field("abort");
+        vd.field("abort");
```
amtep, line 39: This line should probably just be removed. You're doing `vd.field_validated_block("abort"` below.

src/imperator/data/culture.rs
```diff
@@ -62,7 +62,6 @@ impl DbKind for CultureGroup {
             let mut vd = Validator::new(block, data);
             vd.unknown_value_fields(|key, value| {
                 data.verify_exists(Item::Ethnicity, key);
```
amtep, line 64: Looking at the script files, I think this should be
```
key.expect_number();
data.verify_exists(Item::Ethnicity, value);
```

src/imperator/data/event_pictures.rs
```diff
@@ -29,14 +30,19 @@ impl DbKind for EventPicture {
...
-        vd.field_validated_block("picture", |block, data| {
-            let mut vd = Validator::new(block, data);
-            vd.field_item("texture", Item::File);
-            vd.field_validated_block("trigger", |b, data| {
-                validate_trigger(b, data, &mut sc, Tooltipped::No);
-            });
+        vd.multi_field_validated("picture", |bv, data| match bv {
+            BV::Value(t) => {
+                let mut vd = Validator::new(block, data);
+                vd.field_item("picture", Item::File);
```
amtep, line 37: Here don't make another `vd`, because you have a value not a block.

You can do `data.verify_exists(Item::File, t)`

src/imperator/data/event_pictures.rs
```diff
@@ -29,14 +30,19 @@ impl DbKind for EventPicture {
...
+            }
+            BV::Block(b) => {
+                let mut vd = Validator::new(block, data);
```
amtep, line 40: Do `BV::Block(block)` here, so that the `vd` creation below picks up the correct block.

I only do `b` if I actually need to refer to both the parent block and the current block. Otherwise using `block` is better, it prevents mistakes.

src/imperator/data/levy_templates.rs
```diff
@@ -20,17 +20,13 @@ impl LevyTemplate {
 }
 
 impl DbKind for LevyTemplate {
-    fn validate(&self, _key: &Token, block: &Block, data: &Everything) {
+    fn validate(&self, key: &Token, block: &Block, data: &Everything) {
```
amtep, line 23: I guess you're still not using `key` here so it should stay as `_key`

src/imperator/data/missions.rs
```diff
@@ -57,6 +57,8 @@ impl DbKind for Mission {
             validate_trigger(b, data, &mut sc, Tooltipped::No);
         });
 
+        // TODO - Any scopes that are saved in the on_potential or on_start get added to the ScopeContext of every other block in the mission tree, except for the potential block for the entire tree.
+        // Need to figure out how to propogate these save scopes to the new blocks
```
amtep, line 61: Propagating the saved scopes is done automatically as long as you re-use the same `sc`. 

If you want the potential block for the entire tree to not have those, then you can either put it first or make it its own `sc`.

amtep (changes requested):
Added some comments :) I think some need to be fixed before merging.

You should also try
```cargo clippy -p imperator-tiger```

The default `cargo clippy` only checks the ck3 mode.

File: src/imperator/data/missions.rs (hunk 6 above)
Comment (dementive) [reply src/imperator/data/missions.rs:61]: Moving the order should fix the issues with the potential for the whole tree. The main problem though is that MissionTree and MissionTreeTask need to be implemented as seperate Item types and the scopes that are saved in the on_start/on_potential blocks of the MissionTree need to get propogated down to the ScopeContext of all the MissionTreeTask Items that are found in the same file as the MissionTree. Getting the scope context from the MissionTree down to the tasks and only doing it for that file is the main thing I can't figure out here.

File: src/validate.rs [hunk 7]
```diff
@@ -108,7 +108,7 @@ pub fn validate_duration(block: &Block, data: &Everything, sc: &mut ScopeContext
 pub fn validate_optional_duration_int(vd: &mut Validator) {
     let mut count = 0;
 
-    for field in &["days", "weeks", "months", "years"] {
+    for field in &["days", "weeks", "months", "years", "duration"] {
```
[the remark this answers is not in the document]
dementive: It is used in all of the add_X_modifier effects for imperator. In practice it functions the exact same as "days" but for some odd reason the imperator devs implemented it as "duration" instead.

dementive:
I still have a few things to finish up here, I want to add all of the missing gui properties and actually implement the Area, Region, and Religion types.

src/validate.rs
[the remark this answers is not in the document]
amtep: Ok :)

Hmm I'd rather not accept `duration` everywhere because it's probably not valid for the other games or even other duration blocks in Imperator. (Might be worth verifying though)

I think it's ok to make a separate function for this. We already have four, one more won't hurt :)

amtep (commented):
I didn't review yet, just updated the comments from the previous review

src/imperator/scopes.rs
```diff
@@ -2,13 +2,12 @@
...
 
-use fnv::FnvHashMap;
-use once_cell::sync::Lazy;
-
 use crate::everything::Everything;
 use crate::helpers::display_choices;
 use crate::item::Item;
 use crate::scopes::{ArgumentValue, Scopes};
+use fnv::FnvHashMap;
+use once_cell::sync::Lazy;
```
amtep, line 10: The previous location was actually correct :)

amtep (changes requested):
Looks good; just a few leftover issues

src/imperator/data/graphical_culture_types.rs
```diff
@@ -20,8 +19,7 @@ impl GraphicalCultureType {
 }
 
 impl DbKind for GraphicalCultureType {
-    fn validate(&self, key: &Token, block: &Block, data: &Everything) {
-        let mut vd = Validator::new(block, data);
+    fn validate(&self, _key: &Token, _block: &Block, _data: &Everything) {
         // Literally nothing
```
amtep, line 23: In this case it's good to have a separate
`Validator::new(block, data);`
that you don't assign to anything, in order to validate that the block indeed contains nothing. The Validator will do that check when it is dropped.

amtep (approved):

src/imperator/data/event_pictures.rs
```diff
@@ -45,3 +46,18 @@ impl DbKind for EventPicture {
         });
     }
 }
+
+pub fn verify_exists_or_empty(data: &Everything, file: &Token, max_sev: Severity) {
+    let file_str = file.as_str();
+    if file_str.is_empty() {
+        return;
+    }
```
amtep, line 54: `as_str` is a zero-cost method so there's no need to cache it in a variable

src/imperator/data/event_pictures.rs
```diff
@@ -45,3 +46,18 @@ impl DbKind for EventPicture {
...
+    data.fileset.mark_used(&file_str.replace("//", "/"));
+    if !data.fileset.exists(file_str) {
+        let msg = format!("file {file_str} does not exist");
+        report(ErrorKey::MissingFile, Item::File.severity().at_most(max_sev))
+            .msg(msg)
+            .loc(file)
+            .push();
```
amtep, line 61: I'd rather not have this duplicate code. If `verify_exists_implied` changes, this would have to change too. Just calling `data.verify_exists(Item::File, file)` is better.

src/imperator/data/provinces.rs
```diff
@@ -64,6 +64,7 @@ impl ImperatorProvinces {
                             || key.is("river_provinces")
                             || key.is("impassable_terrain")
                             || key.is("uninhabitable")
+                            || key.is("wasteland")
                             || key.is("lakes")
                             || key.is("LAKES")
                         {
```
amtep, line 70: Does this mean they should all really be key.lowercase_is ? To do a case insensitive comparison

src/imperator/tables/modifs.rs
```diff
@@ -101,6 +101,8 @@ pub fn lookup_modif(name: &Token, data: &Everything, warn: Option<Severity>) ->
     if let Some(part) = name_lc.strip_suffix("_combat_bonus") {
         // This is tricky because both Unit and Terrain can have `_` in them.
         // Try each possible separation point in turn.
+        // TODO - This is throwing a false positive for "desert_hills_combat_bonus", but not sure how to fix right now
+        // Both "desert" and "hills" are terrain types...so it is matching hills as terrain and expecting desert to be a unit...
```
amtep, line 105: One thing that might work is to change `part.rmatch_indices` below to `part.match_indices` so that it starts with the first underscore instead of the last.

amtep (changes requested):

src/imperator/data/provinces.rs
dementive: Yeah that would probably work better good call

src/imperator/data/event_pictures.rs
dementive: The reason I added a new function for this is because using `data.verify_exists(Item::File, file)` doesn't really work here because the `icon` and `picture` fields/blocks of both event pictures and event themes allow for the field to be blank but should also still validate if the value is a valid file if it is not blank, there are several hundred instances of them being blank in Invictus so I had to roll my own slightly modified version of this to make it work. Would there be a better way to handle this? How I have it now works but I agree the duplicate code is not ideal.

amtep: Yeah I meant that you can call `data.verify_exists(Item::File, file)` in the new function :) After checking if it's empty

src/imperator/data/decisions.rs
```diff
@@ -1,48 +1,91 @@
...
-#[derive(Clone, Debug)]
-pub struct Decision {}
+#[derive(Clone, Debug, Default)]
+pub struct Decisions {
+    decisions: FnvHashMap<&'static str, Decision>,
+}
+
+impl Decisions {
+    pub fn load_item(&mut self, key: &Token, block: &Block) {
+        if key.is("country_decisions") {
```
amtep, line 26: You could warn here if the key is something unexpected

amtep (dismissed):

src/imperator/data/setup_main.rs
```diff
@@ -60,10 +47,13 @@ fn validate_treasures(block: &Block, data: &Everything) {
         let mut vd = Validator::new(block, data);
         for (_, block) in vd.integer_blocks() {
             let mut vd = Validator::new(block, data);
+            // TODO - This key field should be saved in the db as a new Treasure Item. The Treasure Item would then need to be added to the `treasure:X` scope
+            // How do I save the key as a new db Item type here??? Not sure how to do this right now.
```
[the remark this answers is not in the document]
dementive: Ah yeah I see what I was doing wrong with it now, I'll probably try to go the Item and DbKind approach for this one when I get to doing setup/main. For now im going to remove all the SetupMain code so this PR can get merged finally and later down the line I'll work on finishing up all the setup validators in a new separate PR.

amtep (approved):

PR #43, opened 2024-02-19
ck3: validate artifact reforge blueprints

amtep:


src/ck3/data/artifacts.rs
```diff
@@ -257,3 +258,45 @@ impl DbKind for ArtifactFeatureGroup {
...
+        vd.field_item("in_type", Item::ArtifactType);
+        vd.field_item("in_visuals", Item::ArtifactVisual);
+
+        vd.req_field("out_type");
+        vd.req_field("out_visuals");
+        vd.field_item("out_type", Item::ArtifactType);
+        vd.field_item("out_visuals", Item::ArtifactVisual);
+
+        vd.field_validated_list("disallowed_modifiers", |token, data| {
+            verify_modif_exists(token, data, ModifKinds::all(), Severity::Warning);
```
my4ng, line 290: Consider changing this to `ModifKinds::Character`

src/ck3/data/artifacts.rs
```diff
@@ -257,3 +258,45 @@ impl DbKind for ArtifactFeatureGroup {
...
+        });
+        vd.field_validated_block("replacement_modifiers", |block, data| {
+            let mut vd = Validator::new(block, data);
+            for field in ARTIFACT_RARITY {
+                // There seems to be no way to verify that these are specifically artifact modifiers
+                vd.field_list_items(field, Item::Modifier);
```
my4ng, line 296: Again, if possible, please verify that these will work on Character

my4ng (changes requested):

src/ck3/data/artifacts.rs
amtep: > Consider changing this to `ModifKinds::Character`

Hmm that's a good point, it's possible that artifact bonuses all have to apply to characters.

src/ck3/data/artifacts.rs
amtep: Yeah that can even be a general upgrade. Modifier's `validate_call` can deduce the `ModifKind` from the scope types of the received `sc` and verify that all the modifer's modifs apply to that modifkind.

amtep: It turned out that `validate_call` doesn't need to do that because the callers already call `validate_property_use` :) So I made the artifact code call that too.

my4ng (approved):

PR #44, opened 2024-02-19
imperator: support hidden:scope:foo

amtep:


src/trigger.rs
```diff
@@ -1016,11 +1016,11 @@ pub fn validate_target_ok_this(
             Part::TokenArgument(func, arg) => validate_argument(part_flags, func, arg, data, sc),
             Part::Token(part) => {
                 // prefixed scope transition, e.g. cp:councillor_steward
-                if let Some((prefix, mut arg)) = part.split_once(':') {
+                if let Some((prefix, mut arg)) = part.split_last_once(':') {
                     // event_id have multiple parts separated by `.`
                     let is_event_id = prefix.lowercase_is("event_id");
                     if is_event_id {
-                        arg = token.split_once(':').unwrap().1;
+                        arg = token.split_last_once(':').unwrap().1;
```
my4ng, line 1023: I don't think this is needed?

my4ng (changes requested):
This looks fine but I am not convinced this warrants using `split_last_once`. Perhaps a check for a special case is more suitable: have `hidden` in the `SCOPE_PREFIX` table and check that the argument starts with `scope` and adjust things accordingly. Will that work?

PR #46, opened 2024-02-22
Local value parsing

my4ng:
- Replace `StringTable` with `Bump` allocator
- Rework parsing, esp. local values


src/parse/pdxfile.rs
```diff
@@ -817,319 +835,525 @@ pub fn parse_pdx_macro(inputs: &[Token], local_macros: LocalMacros) -> Block {
...
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+/// Kinds of [`MacroComponent`].
+pub enum MacroComponentKind {
+    Source,
+    LocalValue,
+    Macro,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+/// Macro components outputted from [`split_macros`].
```
amtep, line 1142: Technically the past perfect of 'output' is 'output' :) It's an irregular verb.

src/parse/pdxfile.rs
```diff
@@ -69,6 +96,75 @@ impl Calculation {
...
+    fn new() -> Self {
+        Self { stack: Vec::new(), current: Vec::new() }
+    }
+
+    fn start(&mut self) {
+        self.stack.clear();
+        self.current.clear();
+    }
+
+    /// Register a part of a calculation, either an operator or a [`Value`](`Calculation::Value`).
```
amtep, line 118: I think this comment was never true :) From the body it's clear that the `op` must be an operator, not a `Value`.

amtep (changes requested):
Some style comments; no bugs found.

Please fix the compiler warning about `old_warn` :) The function can just be removed now.

src/block/blockitem.rs
```diff
@@ -15,7 +15,7 @@ impl BlockItem {
             Some(field)
         } else {
             let msg = format!("unexpected {}", self.describe());
-            error_info(self, ErrorKey::Structure, &msg, "Did you forget an = ?");
+            err(ErrorKey::Structure).msg(msg).info("Did you forgetr an = ?").loc(self).push();
```
amtep, line 18: Typo :)

src/block/blockitem.rs
```diff
@@ -25,7 +25,7 @@ impl BlockItem {
...
```
amtep, line 28: Here too

src/block.rs
```diff
@@ -394,51 +394,53 @@ impl Block {
...
+                                val.loc.link_idx = Some(MACRO_MAP.get_or_insert_loc(orig_loc));
+                                content.push(val);
+                                break;
+                            }
                         }
                     }
                 }
+                //  else {
+
+                // }
```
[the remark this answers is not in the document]
my4ng: that was just remnant of the if-else from earlier

src/pdxfile.rs
```diff
@@ -25,6 +27,21 @@ pub enum PdxEncoding {
     Detect,
 }
 
+fn strip_bom(contents: String) -> String {
```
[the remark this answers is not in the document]
my4ng: I think the last warning is most likely correct, but I just want to confirm if my understanding is correct:
The global allocator does not have the offset-ed pointer in its record, so likely cause an UB when trying to drop the returned String.

src/parse/pdxfile.rs
```diff
@@ -28,10 +41,12 @@ enum State {
...
     Comparator,
+    /// Parsing a macro surrounded by `$...$`.
+    Macro,
+    /// Parsing a local value `@...`
+    LocalValue,
     /// Parsing a `@[ ... ]` local value calculation.
-    Calculation,
-    /// Parsing a local value id in a `Calculation`.
-    CalculationId,
+    Calculation(Option<(usize, Loc)>),
```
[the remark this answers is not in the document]
my4ng: That's a good point, thanks! Yeah best for uniformity

src/parse/pdxfile.rs
```diff
@@ -559,254 +595,236 @@ impl Parser {
...
                     }
                 }
                 State::Comparator => {
                     if c.is_comparator_char() {
-                        current_id.push(c);
+                        current_id.push(c, content, i, loc);
                     } else {
-                        parser.comparator(&take(&mut current_id), token_start);
+                        let token = current_id.token();
+                        parser.comparator(token.as_str(), token.loc);
```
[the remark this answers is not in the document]
my4ng: I am hesitant to do so since it mostly uses the str and loc separately

src/parse/pdxfile.rs
my4ng: Right now, it does not forbid taking a Value variant. Should there be a check for it do you think?

amtep: I think it's OK to just fix the comment.

I realize it was my comment and not yours but it would be convenient to fix it here :)

src/pdxfile.rs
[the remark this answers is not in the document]
amtep: > The global allocator does not have the offset-ed pointer in its record, so likely cause an UB when trying to drop the returned String.
Yeah exactly

amtep (approved):

PR #47, opened 2024-02-22
ck3: validate lease contracts

amtep:


src/ck3/data/lease_contracts.rs
```diff
@@ -0,0 +1,86 @@
...
+impl LeaseContract {
+    pub fn add(db: &mut Db, key: Token, block: Block) {
+        db.add(Item::LeaseContract, key, block, Box::new(Self {}));
+    }
+}
+
+impl DbKind for LeaseContract {
+    fn validate(&self, key: &Token, block: &Block, data: &Everything) {
+        let mut vd = Validator::new(block, data);
+        vd.req_field("hierarchy");
```
my4ng, line 32: > "Definition for automatic lease hierarchy. Required for some built-in lease contracts, not supported for the others."
best to only require and allow for the built-in ones

src/ck3/data/lease_contracts.rs
```diff
@@ -0,0 +1,86 @@
...
+                validate_trigger(block, data, &mut sc, Tooltipped::No);
+            });
+            vd.field_validated_key_block("barony_valid", |key, block, data| {
+                let mut sc = ScopeContext::new(Scopes::LandedTitle, key);
+                validate_trigger(block, data, &mut sc, Tooltipped::No);
+            });
+            {
+                let mut sc = ScopeContext::new(Scopes::Character, key);
+                vd.field_target("lessee", &mut sc, Scopes::Character);
+            }
```
my4ng, line 51: Not sure why it is enclosed by braces. Perhaps can reuse sc in other field validations too

src/ck3/data/lease_contracts.rs
```diff
@@ -0,0 +1,86 @@
...
+        });
+
+        vd.field_item("government", Item::GovernmentType);
```
my4ng, line 54: undocumented comment

src/ck3/data/lease_contracts.rs
```diff
@@ -0,0 +1,86 @@
...
+                        let mut sc = ScopeContext::new(Scopes::None, key);
+                        sc.define_name("ruler", Scopes::Character, key);
+                        sc.define_name("lessee", Scopes::Character, key);
+                        vd.field_validated_block_sc(
+                            "weight",
+                            &mut sc,
+                            validate_modifiers_with_base,
+                        );
+                        vd.field_choice("beneficiary", &["ruler", "lessee"]);
+                        vd.field_choice("rest", &["ruler", "lessee"]);
```
my4ng, line 80: Should/Must rest be the other one from beneficiary or not?

src/ck3/data/lease_contracts.rs
```diff
@@ -0,0 +1,86 @@
...
+                vd.field_target("lessee", &mut sc, Scopes::Character);
+            }
+        });
+
+        vd.field_item("government", Item::GovernmentType);
+        vd.field_list_items("valid_holdings", Item::HoldingType);
+        vd.field_integer("ruler_share_min_opinion_from_lessee");
+        vd.field_choice("hook_strength_max_opinion", &["none", "any", "strong"]);
+
+        for field in &["tax", "levy"] {
```
my4ng, line 59: > "Note: Levies can only go to one character on each level. This means levies cannot be assigned in part to the ruler and in part to the lessee because they are considered to be on the same level. So you can use only 0 or 100 when splitting between ruler and lessee."

my4ng (changes requested):

src/ck3/data/lease_contracts.rs
amtep: Oh I see

I don't read the .infos carefully enough

src/ck3/data/lease_contracts.rs
amtep: Lol I didn't read past "See tax for syntax"

src/ck3/data/lease_contracts.rs
amtep: Yeah it's deliberate, to avoid re-using it later.

My habit of re-using sc's in other validators is a bit messy, since they do remember scopes that are set in each use.

However you can't set scopes in a target field so there's no need for the braces here :)

src/ck3/data/lease_contracts.rs
amtep: Honestly it's weird that they support both

src/ck3/data/lease_contracts.rs
amtep: I'm not sure this comment is true. The existing `theocracy_lease` doesn't follow it.

amtep: I also remember from playing that you get different amounts of troops from the chaplain depending on his opinion of you

my4ng: Yes from the look of it it seems that comment is outdated since clearly they don't follow that as you pointed out.

my4ng (approved):

PR #49, opened 2024-02-23
vic3: validate the new combat units

amtep:


src/vic3/data/combat_units.rs
```diff
@@ -46,6 +48,101 @@ impl DbKind for OldCombatUnit {
...
+        });
+        vd.field_validated_key_block("can_build_conscript", |key, block, data| {
+            let mut sc = ScopeContext::new(Scopes::Country, key);
+            validate_trigger(block, data, &mut sc, Tooltipped::No);
+        });
+        vd.field_list_items("unlocking_technologies", Item::Technology);
+        let mut seen_unconditional = false;
+        vd.multi_field_validated_key_block("combat_unit_image", |key, block, data| {
+            if seen_unconditional {
+                let msg = "there was a previous `combat_unit_image` without a trigger, so this one will not be used";
```
my4ng, line 98: would be better if can provide the location of that previous block saved in `seen_unconditional: Option<Token/Block>` and then point to it with `loc(seen_block, "from here")`. Also consider suggesting adding something like "; move that to the end as the fallback".

src/vic3/data/combat_units.rs
```diff
@@ -46,6 +48,101 @@ impl DbKind for OldCombatUnit {
...
+
+impl CombatUnitGroup {
+    pub fn add(db: &mut Db, key: Token, block: Block) {
+        db.add(Item::CombatUnitGroup, key, block, Box::new(Self {}));
+    }
+}
+
+impl DbKind for CombatUnitGroup {
+    // This whole item type is undocumented.
+    fn validate(&self, key: &Token, block: &Block, data: &Everything) {
```
my4ng, line 134: Also verify loca for it and +desc

my4ng (changes requested):

src/vic3/data/ai_strategies.rs
```diff
@@ -85,10 +107,16 @@ impl DbKind for AiStrategy {
...
+        vd.field_script_value_rooted("wanted_construction_output", Scopes::Country);
+        vd.replaced_field("wanted_construction_sector_levels", "wanted_construction_output");
         vd.field_script_value_rooted("wanted_army_size", Scopes::Country);
         vd.field_script_value_rooted("wanted_navy_size", Scopes::Country);
 
+        vd.field_validated_key_block(
+            "combat_unit_group_weights",
+            validate_combat_unit_group_weights,
+        );
+        vd.field_script_value_rooted("conscript_battalion_ratio", Scopes::Country);
```
my4ng, line 119: "scope:military_formation = military formation"

my4ng (changes requested):

PR #52, opened 2024-02-23
vic3: validate alerts

amtep:


src/vic3/data/alerts.rs
```diff
@@ -0,0 +1,125 @@
...
+        let loca = format!("alert_{key}_action");
+        data.verify_exists_implied(Item::Localization, &loca, key);
+        let loca = format!("{key}_setting_name");
+        data.verify_exists_implied(Item::Localization, &loca, key);
+
+        vd.req_field("script_context");
+        let mut sc = ScopeContext::new(Scopes::None, key);
+        if let Some(script_context) = vd.field_value("script_context") {
+            let mut found = false;
+            for &(context, s) in SCRIPT_CONTEXTS {
```
my4ng, line 47: Might be better as a FnvHashMap for uniformity but not necessary

my4ng (approved):

src/vic3/data/alerts.rs
amtep: I considered it but this array is used so rarely, comparatively, that it didn't seem worth the extra code

PR #56, opened 2024-02-28
ck3: validate map_data/climate.txt

amtep:


src/everything.rs
```diff
@@ -454,6 +455,7 @@ impl Everything {
         s.spawn(|_| self.events_ck3.validate(self));
         s.spawn(|_| self.provinces_ck3.validate(self));
         s.spawn(|_| self.wars.validate(self));
+        s.spawn(|_| Climate::validate_all(&self.database, self));
```
my4ng, line 458: Isn't it validated below in `self.database.validate(self);`?

my4ng (changes requested):

src/everything.rs
amtep: I guess the name is confusing :) This call is for a general validation of the item type, instead of per item.

amtep: Maybe just `validate_item_type`

my4ng (approved):

PR #59, opened 2024-02-29
Province terrain and properties validation

my4ng:


src/ck3/data/prov_terrain.rs
```diff
@@ -0,0 +1,172 @@
...
+use crate::report::{warn, ErrorKey};
+use crate::token::Token;
+use crate::validator::Validator;
+use crate::Severity;
+
+use super::provinces::ProvId;
+
+#[derive(Clone, Debug, Default)]
+pub struct ProvinceTerrains {
+    provinces: FnvHashSet<ProvinceTerrain>,
```
amtep, line 21: I'd like a comment here explaining that ProvinceTerrain is only hashed by its key. Otherwise this HashSet looks weird.

src/ck3/data/provinces.rs
```diff
@@ -128,12 +128,19 @@ impl Ck3Provinces {
...
     }
 
+    pub(crate) fn verify_exists_provid(&self, provid: ProvId, item: &Token, max_sev: Severity) {
+        if !self.provinces.contains_key(&provid) {
+            let msg = format!("province {provid} not defined in map_data/definition.csv");
+            report(ErrorKey::MissingItem, Item::Province.severity().at_most(max_sev))
+                .msg(msg)
+                .loc(item)
+                .push();
+        }
```
amtep, line 138: The usual way to do this is to call something like `self.provinces.verify_exists(provid)` (possibly passing max_sev in too) so that `self.provinces` can generate the message about it not existing. That way the knowledge about `map_data/definition.csv` can stay encapsulated there.
(Especially since the logic of what makes a provid valid in the game is not 100% known yet)

The interface between `Everything` and the `FileHandler`s works that way too.

amtep (changes requested):
Looks good :) But still some notes.

src/ck3/data/prov_terrain.rs
my4ng: No `ProvId` is used here as a hashmap key because there are also `default_*` ones at the top, which obviously cannot be parsed to ProvId`, but I do like to verify those too. I will add this to the comment.

src/ck3/data/provinces.rs
my4ng: Yeah the reason why I didn't use that is that normally verify_exists requires a token as the argument but here we are using `ProvId` so I don't want to be confusing. Since we have already parsed the ProvId already, there isn't any point to still pass the token and parse all over again, especially we have verified that it won't generate an error.

src/ck3/data/prov_terrain.rs
```diff
@@ -10,32 +9,49 @@ use crate::helpers::dup_error;
...
-        let province = ProvinceTerrain::new(key, value);
-        if let Some(existing_province) = self.provinces.get(&province) {
-            if existing_province.key.loc.kind >= province.key.loc.kind {
-                dup_error(&province.key, &existing_province.key, "province");
+    fn load_item(&mut self, id: ProvId, key: Token, value: Token) {
+        if let Some(province) = self.provinces.get_mut(&id) {
+            if province.key.loc.kind >= key.loc.kind {
+                dup_error(&key, &province.key, "province");
             }
+            province.terrain = value;
```
amtep, line 33: I don't think this is enough, because now the province.key token has the wrong loc.

src/ck3/data/prov_terrain.rs
```diff
@@ -10,32 +9,49 @@ use crate::helpers::dup_error;
...
-            item.validate(data);
+        for (provid, item) in &self.provinces {
+            item.validate(*provid, data);
+        }
+
+        for (name, terrains) in DEFAULT_TERRAINS.iter().zip(&*self.defaults) {
+            if let Some(terrain) = terrains {
+                data.verify_exists(Item::Terrain, terrain);
+            } else {
+                let msg = format!("missing default terrain: {}", name);
```
amtep, line 48: Clippy complaint :)

src/ck3/data/prov_terrain.rs
```diff
@@ -10,32 +9,49 @@ use crate::helpers::dup_error;
...
 use super::provinces::ProvId;
 
+const DEFAULT_TERRAINS: &[&str] = &["default_land", "default_sea", "default_coastal_sea"];
+
 #[derive(Clone, Debug, Default)]
 pub struct ProvinceTerrains {
-    provinces: FnvHashSet<ProvinceTerrain>,
+    provinces: FnvHashMap<ProvId, ProvinceTerrain>,
+    file_loc: Option<Loc>,
+    defaults: Box<[Option<Token>; DEFAULT_TERRAINS.len()]>,
```
amtep, line 24: Is the Box needed? A constant-size array can be declared directly I think

amtep (changes requested):

src/ck3/data/prov_terrain.rs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        for (name, terrains) in DEFAULT_TERRAINS.iter().zip(&*self.defaults) {
+            if let Some(terrain) = terrains {
+                data.verify_exists(Item::Terrain, terrain);
+            } else {
+                let msg = format!("missing default terrain: {}", name);
+                warn(ErrorKey::Validation)
+                    .msg(msg)
+                    // SAFETY: `file_loc` has been set in `handle_file`.
```
amtep, line 51: Turns out this panics if there were no province terrain files at all.

It runs into a general problem with tiger, which is that there's not always a good Loc to use for a missing item.

Maybe a synthetic FileEntry for the whole directory would be a good default. FileEntry can be used as an errorloc and can contain arbitrary paths, unlike Loc which has a PathTableIndex.

my4ng: Yes that's the problem. Of course there is the FileEntry from `handle_file` but that faces the same problem right now. I wonder if it is possible to have a Static Loc that can be used for all missing items.

amtep (approved):

PR #61, opened 2024-03-01
vic3: validate scripted rules

amtep:


src/vic3/data/scripted_rules.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Debug, Clone)]
+struct ScriptedRuleScopeContext {
+    tooltipped: Tooltipped,
+    root: Scopes,
+    names: Vec<(String, Scopes)>,
+    lists: Vec<(String, Scopes)>,
+}
+
+/// Processed version of [`SCRIPTED_RULES`].
+static SCRIPTED_RULE_SCOPES_MAP: Lazy<FnvHashMap<String, ScriptedRuleScopeContext>> =
```
my4ng, line 60: Would it be better to simply save them as `ScriptedRuleScopeContext`s rather than being parsed?

amtep: The parsing code was copied from ck3 :) I think we'll be glad of the flexibility if more rules are added

amtep: Unifying the parsing machinery might be nice but I'm not sure how

my4ng (approved):

PR #63, opened 2024-03-02
store gui widget names as items

amtep:
This will be useful for some validators

### DIFF
--- a/src/data/gui.rs
+++ b/src/data/gui.rs
@@ -21,17 +21,17 @@ use crate::validator::Validator;
 #[derive(Debug, Default)]
 pub struct Gui {
     files: FnvHashMap<PathBuf, Vec<GuiWidget>>,
-    templates: FnvHashMap<String, GuiTemplate>,
+    templates: FnvHashMap<&'static str, GuiTemplate>,
     // Type keys are stored in lowercase because type lookup is case-insensitive
-    types: FnvHashMap<String, GuiType>,
-    layers: FnvHashMap<String, GuiLayer>,
+    types: FnvHashMap<Lowercase<'static>, GuiType>,
+    layers: FnvHashMap<&'static str, GuiLayer>,
     // TextIcon is in a Vec because a single icon can have multiple definitions with different
     // iconsize parameters.
-    texticons: FnvHashMap<String, Vec<TextIcon>>,
-    textformats: FnvHashMap<String, TextFormat>,
+    texticons: FnvHashMap<&'static str, Vec<TextIcon>>,
+    textformats: FnvHashMap<&'static str, TextFormat>,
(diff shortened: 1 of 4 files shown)


src/data/gui.rs
```diff
@@ -74,6 +75,9 @@ impl Gui {
                 }
             }
         } else {
+            if let Some(name) = block.get_field_value("name") {
+                self.widget_names.insert(name.to_string(), name.clone());
```
my4ng, line 79: Can replace with `name.as_str()` and use `&'static str`

src/data/gui.rs
```diff
@@ -227,6 +231,14 @@ impl Gui {
         self.textformats.values().map(|item| &item.key)
     }
 
+    pub fn name_exists(&self, key: &str) -> bool {
+        self.widget_names.contains_key(key)
+    }
+
+    pub fn iter_names(&self) -> impl Iterator<Item = &Token> {
+        self.widget_names.values()
```
my4ng, line 239: Would it be better to use a hashset and a vec instead? Since we don't need to retrieve the mapped value, it would improve the cache hits for `iter_names`.

my4ng (changes requested):

src/data/gui.rs
amtep: I think it's not worth optimizing since iter_names is only used by the proof-of-concept item dumper

src/data/gui.rs
amtep: Will do

src/data/gui.rs
```diff
@@ -21,17 +21,17 @@ use crate::validator::Validator;
...
     // iconsize parameters.
-    texticons: FnvHashMap<String, Vec<TextIcon>>,
-    textformats: FnvHashMap<String, TextFormat>,
+    texticons: FnvHashMap<&'static str, Vec<TextIcon>>,
+    textformats: FnvHashMap<&'static str, TextFormat>,
     // This is indexed by a (colorblindmode, textformatname) pair
-    textformats_colorblind: FnvHashMap<(String, String), TextFormat>,
-    widget_names: FnvHashMap<String, Token>,
+    textformats_colorblind: FnvHashMap<(&'static str, &'static str), TextFormat>,
+    widget_names: FnvHashMap<&'static str, Token>,
```
my4ng, line 34: Probably can use Hashset now that Token implements Borrow<str>

my4ng (changes requested):

src/data/gui.rs
amtep: done :)  I had to rebase to get the Borrow implementation in

amtep:
I rebased to fix the merge conflict

my4ng (approved):

PR #65, opened 2024-03-02
vic3: handle replace_paths

amtep:
Also fix vic3-tiger-auto which still had a lot of ck3 code

### DIFF
--- a/src/fileset.rs
+++ b/src/fileset.rs
@@ -17,6 +17,9 @@ use crate::everything::Everything;
 use crate::everything::FilesError;
 use crate::game::Game;
 use crate::item::Item;
+#[cfg(feature = "vic3")]
+use crate::mod_metadata::ModMetadata;
+#[cfg(any(feature = "ck3", feature = "imperator"))]
 use crate::modfile::ModFile;
 use crate::pathtable::{PathTable, PathTableIndex};
 use crate::report::{
@@ -264,6 +267,7 @@ impl Fileset {
             let label =
                 block.get_field_value("label").map_or_else(default_label, ToString::to_string);
             if Game::is_ck3() || Game::is_imperator() {
+                #[cfg(any(feature = "ck3", feature = "imperator"))]
                 if let Some(path) = block.get_field_value("modfile") {
                     let path = PathBuf::from(path.as_str());
                     if let Ok(modfile) = ModFile::read(&path) {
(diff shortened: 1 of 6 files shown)


src/mod_metadata.rs
```diff
@@ -0,0 +1,52 @@
...
+        if let Some(custom_data) = self.block.get_field_block("game_custom_data") {
+            if let Some(replace_paths) = custom_data.get_field_list("replace_paths") {
+                return replace_paths.iter().map(|t| PathBuf::from(t.as_str())).collect();
+            }
+        }
+        Vec::new()
+    }
+
+    /// The mod's name in human-friendly form, if available.
+    pub fn display_name(&self) -> Option<String> {
```
my4ng, line 49: Doesn't need to return Option<String>. Option<&'static str> suffices.

src/parse/json.rs
```diff
@@ -220,6 +262,14 @@ fn parse(blockloc: Loc, content: &str) -> Block {
             err(ErrorKey::ParseError).msg("Quoted string not closed").loc(&token).push();
             parser.token(token);
         }
+        State::Id => {
+            let token = Token::new(&current_id, token_start);
+            if token.is("true") || token.is("false") {
+                parser.token(token);
+            } else {
+                warn(ErrorKey::ParseError).msg("unexpected unquoted string").loc(token).push();
```
my4ng, line 270: maybe with info that can only be either true or false?

src/parse/json.rs
```diff
@@ -190,6 +197,41 @@ fn parse(blockloc: Loc, content: &str) -> Block {
...
+            State::Id => {
+                if c.is_alphabetic() {
+                    current_id.push(c);
+                } else {
+                    let token = Token::new(&take(&mut current_id), token_start);
+                    if token.is("true") || token.is("false") {
+                        parser.token(token);
+                    } else {
+                        warn(ErrorKey::ParseError)
+                            .msg("unexpected unquoted string")
```
my4ng, line 209: same as below

src/mod_metadata.rs
```diff
@@ -0,0 +1,52 @@
...
+        let in_mod_path = PathBuf::from(".metadata/metadata.json");
+        let pathname = fix_slashes_for_target_platform(mod_dir.join(&in_mod_path));
+        let entry = FileEntry::new(in_mod_path, FileKind::Mod, pathname.clone());
+        let block = parse_json_file(&entry)
+            .with_context(|| format!("Could not read metadata file {}", pathname.display()))?;
+        Ok(Self { modpath: mod_dir.to_path_buf(), block })
+    }
+
+    /// Return the full path to the mod root.
+    pub fn modpath(&self) -> PathBuf {
```
my4ng, line 34: can change to &Path return value

my4ng (changes requested):

src/fileset.rs
```diff
@@ -264,6 +267,7 @@ impl Fileset {
             let label =
                 block.get_field_value("label").map_or_else(default_label, ToString::to_string);
             if Game::is_ck3() || Game::is_imperator() {
+                #[cfg(any(feature = "ck3", feature = "imperator"))]
```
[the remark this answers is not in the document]
amtep: It's needed because of the references to `ModFile` which isn't defined for vic3.

I guess it does make the `if` redundant

src/mod_metadata.rs
amtep: Ah yeah. This was copied from the signature of `ModFile::modpath`, which may have to construct a new PathBuf in its body.

src/fileset.rs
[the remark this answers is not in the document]
my4ng: Mhmm actually not if its got ck3 and vic3 feature and is actually vic3 game then wont work :)

my4ng (approved):

my4ng (approved):

PR #67, opened 2024-03-02
report missing name loca as untidy in some cases

amtep:


src/data/localization.rs
```diff
@@ -285,6 +285,41 @@ impl Localization {
         }
     }
 
+    #[cfg(feature = "ck3")]
+    pub fn verify_name_exists(&self, name: &Token, max_sev: Severity) {
+        if name.as_str().is_empty() {
+            return;
```
my4ng, line 291: Should empty name return an error?

src/data/localization.rs
```diff
@@ -285,6 +285,41 @@ impl Localization {
...
+                langs.push(lang);
+            }
+        }
+        if !langs.is_empty() {
+            // It's merely untidy if the name is only missing in latin-script languages and the
+            // name doesn't have indicators that it really needs to be localized (such as underscores
+            // or extra uppercase letters). In all other cases it's a warning.
+            let sev = if only_latin_script(&langs)
+                && !name.as_str().contains('_')
+                && name.as_str().chars().filter(|c| c.is_uppercase()).count() == 1
```
my4ng, line 307: Names with multiple parts may contain more than one uppercase chars

src/data/localization.rs
```diff
@@ -285,6 +285,41 @@ impl Localization {
...
+            let hash = self.locas.get(lang);
+            if hash.is_none() || !hash.unwrap().contains_key(name.as_str()) {
+                langs.push(lang);
+            }
+        }
+        if !langs.is_empty() {
+            // It's merely untidy if the name is only missing in latin-script languages and the
+            // name doesn't have indicators that it really needs to be localized (such as underscores
+            // or extra uppercase letters). In all other cases it's a warning.
+            let sev = if only_latin_script(&langs)
```
my4ng, line 305: This is to assume that name only contains latin chars, which may not be the case. For example, if the name is cyrillic and the mod only supports Russian, then this will be warning still, even though it doesn't need to be. On the other hand, if name is Chinese characters, and missing Latin locas, this would only be untidy. Simply checking ascii doesn't work either, since there could be accented characters.

An improvement could be to use unicode blocks to guess the language, e.g. [unicode_blocks](https://docs.rs/unicode-blocks/latest/unicode_blocks/).

my4ng (changes requested):

src/data/localization.rs
amtep: Hmm yeah. It's built around the assumption that the input name is English

src/data/localization.rs
amtep: Yeah probably. I copied this from the more general verify check, where it makes more sense.

src/data/localization.rs
amtep: Yeah but in practice those (should) contain _ so that they're localization. Having a space in a name is actually a greater sin because then you can't localize it at all :)

my4ng: Unfortunately, there are names connected with hyphen, like "Al-Nasir"

my4ng (approved):

PR #73, opened 2024-03-05
Optimization

my4ng:
Benchmark result (CK3-tiger):
- 0-2% for small mods
- ~15% for large mods


src/ck3/data/provinces.rs
```diff
@@ -15,10 +16,43 @@ use crate::token::{Loc, Token};
 
 pub type ProvId = u32;
 
-#[derive(Clone, Debug, Default)]
+const COLOUR_COUNT: usize = 256 * 256 * 256;
+#[derive(Clone, Copy, Debug, Default)]
```
amtep, line 20: Should it really be Copy? This is a 2MB array, right?

src/datatype.rs
```diff
@@ -749,201 +752,140 @@ fn lookup_global_promote(lookup_name: &str) -> Option<(Args, Datatype)> {
...
-    lookup_name: &str,
-    ltype: Datatype,
-    global: &[(&str, Datatype, Args, Datatype)],
-) -> LookupResult {
-    let start = global.partition_point(|(name, _, _, _)| name < &lookup_name);
-    let mut found_any = false;
+#[inline]
+fn lookup_global_function(lookup_name: &str) -> Option<(Args, Datatype)> {
+    global_function(lookup_name)
+}
```
amtep, line 770: Why these tiny functions? I don't understand why `lookup_global_function` and `global_function` are separate

src/datatype.rs
```diff
@@ -749,201 +752,140 @@ fn lookup_global_promote(lookup_name: &str) -> Option<(Args, Datatype)> {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.0.eq_ignore_ascii_case(other.0)
+    }
+}
+
+impl Eq for CaseInsensitiveStr {}
+
+impl std::hash::Hash for CaseInsensitiveStr {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.0.to_ascii_lowercase().hash(state);
```
amtep, line 848: This might be an expensive hash since it allocates a String. How often is it called?

amtep (commented):
Just two questions :)

src/datatype.rs
my4ng: This is purely to mirror promotes and functions, but yeah i think they should be removed

src/datatype.rs
my4ng: This is called for every insertion and every query. I don't think there is anyway to avoid this unfortunately.

amtep (approved):

PR #75, opened 2024-03-06
bugfix: support case-insensitive modif lookups

amtep:


amtep:
I considered pushing the use of Lowercase further up (to name_lc), but then I have to implement strip_prefix and strip_suffix and I can't decide on the API for those. Or rather, the proper API would have vastly expanded the line lengths in lookup_modif()

amtep:
I added a CaseFolded type that follows Unicode rules and used that instead of Lowercase.

amtep:
Turns out the game engine only does ascii lowercase, so using case folding was overkill

src/ck3/data/traits.rs
```diff
@@ -27,6 +28,11 @@ pub struct Traits {
     tracks: FnvHashSet<Token>,
     constraints: FnvHashSet<Token>,
     flags: FnvHashSet<Token>,
+
+    // Casefolded registries of the above collections, for case-insensitive lookups
```
my4ng, line 32: Should be lowercased now

src/ck3/data/traits.rs
```diff
@@ -27,6 +28,11 @@ pub struct Traits {
...
+    traits_lc: FnvHashMap<Lowercase<'static>, &'static str>,
```
my4ng, line 33: Will CaseInsensitiveStr set be better here?

src/db.rs
```diff
@@ -26,14 +27,18 @@ pub struct Db {
     /// Items generated as side effects of the full items in `database`.
     /// The `Vec` is indexed with an `Item` discriminant.
     flags: Vec<FnvHashSet<Token>>,
+    /// Case-folded registry of database items and flags, for case insensitive lookups
```
my4ng, line 30: Same as before should now be lowercased

src/db.rs
```diff
@@ -110,6 +122,22 @@ impl Db {
         }
     }
 
+    #[cfg(feature = "ck3")] // vic3 happens not to use
+    pub fn cf_has_property(
```
my4ng, line 126: should be lc_*

src/lowercase.rs
```diff
@@ -16,10 +23,10 @@ impl<'a> Lowercase<'a> {
     /// Take a string and return the lowercased version.
     pub fn new(s: &'a str) -> Self {
         // Avoid allocating if it's not necessary
-        if s.chars().all(char::is_lowercase) {
-            Lowercase(Cow::Borrowed(s))
+        if s.chars().any(|c| c.is_ascii_uppercase()) {
```
my4ng, line 26: does char::is_ascii_uppercase work here too?

src/lowercase.rs
```diff
@@ -44,12 +51,33 @@ impl<'a> Lowercase<'a> {
...
+    }
+
+    #[cfg(any(feature = "vic3", feature = "imperator"))]
+    pub fn rmatch_indices_unchecked(&self, separator: char) -> RMatchIndices<char> {
+        self.0.rmatch_indices(separator)
+    }
+
+    /// Implementing `SliceIndex` directly requires some nightly-only API
+    #[cfg(any(feature = "vic3", feature = "imperator"))]
+    pub fn slice<R: 'a + SliceIndex<str, Output = str>>(&'a self, range: R) -> Self {
```
my4ng, line 79: you can implement [std::ops::Index](https://doc.rust-lang.org/std/ops/trait.Index.html)

my4ng (changes requested):
Just some minor things otherwise all good

src/lowercase.rs
amtep: No it complained about a missing `for<'a>` and I didn't understand its problem

src/lowercase.rs
amtep: Aha I didn't know that worked on ranges

src/db.rs
amtep: I was searching for `_cf` and missed it :)

src/lowercase.rs
my4ng: yeah i've encountered that too... surprising since it basically worked in `all(char::is_ascii_lowercase)`

src/lowercase.rs
amtep: I tried but I can't get it to work because the trait method returns a reference. I can construct a `Lowercase` to return, but I can't return a reference to one I construct inside the function.

my4ng (approved):

amtep:
Merged separately

PR #83, opened 2024-03-09
Fix a race condition in the macro map

amtep:


amtep:
Going to restore ::clear() for use by benchmarks

PR #84, opened 2024-03-10
ck3: check that every epidemic has a deathreason

amtep:


src/ck3/data/deathreasons.rs
```diff
@@ -22,6 +22,11 @@ impl DeathReason {
     pub fn add(db: &mut Db, key: Token, block: Block) {
         db.add(Item::DeathReason, key, block, Box::new(Self {}));
     }
+
+    #[allow(clippy::unused_self)] // whether self is used should not affect the API here
+    pub fn is_for_epidemic(&self, _key: &Token, block: &Block, epidemic_key: &str) -> bool {
+        block.field_value_is("epidemic", epidemic_key)
+    }
```
my4ng, line 29: will caching the value in self be better?

amtep: Hmm maybe. It's currently about 10 epidemics * about 100 death reasons * 10 fields each = 10,000 ops. Might be worth reducing.

my4ng (approved):

PR #86, opened 2024-03-12
misc bugfixes

amtep:


src/effect.rs
```diff
@@ -452,28 +460,18 @@ pub fn validate_effect_control(
...
-            });
-        });
+        // Mark all these as known fields. They will be validated after the main loop,
+        // because the main loop may set named scopes for them to use.
+        vd.field("title");
+        vd.field("desc");
+        vd.field("tooltip");
+        vd.field("left_icon");
+        vd.field("right_icon");
+        vd.field("localization_values");
```
my4ng, line 474: Is it necessary to be known fields since they are checked later?

src/vic3/tables/triggers.rs
```diff
@@ -426,7 +426,7 @@ const TRIGGER: &[(Scopes, &str, Trigger)] = &[
     (Scopes::Country, "has_wasted_construction", Boolean),
     (Scopes::None, "hidden_trigger", Control),
     (Scopes::Country, "highest_secession_progress", CompareValue),
-    (Scopes::InterestGroup, "ig_approval", CompareChoice(APPROVALS)),
+    (Scopes::InterestGroup, "ig_approval", CompareApproval),
```
my4ng, line 429: Mhmmm I originally made the change to move away from specialization. I think a `CompareNumberChoice` would be more appropriate for code reuse later.

src/effect.rs
amtep: Yes because the main loop traverses unknown fields. Marking them here will except them from regular effects processing.

src/vic3/tables/triggers.rs
amtep: It's a good point but we do have only one user for it :)

I'll change it though

src/effect.rs
amtep: I clarified the comment in the rebase

src/vic3/tables/triggers.rs
amtep: Done

PR #90, opened 2024-03-19
Misc. optimizations

my4ng:
- Use `to_ascii_lowercase`
- Rename `Id` to `Cob` and use in loca. parsing


src/parse/localization.rs
```diff
@@ -366,6 +368,7 @@ pub struct ValueParser<'a> {
     content_iters: Vec<Peekable<Chars<'a>>>,
     content_idx: usize,
     value: Vec<LocaValue>,
+    text: Cob,
```
amtep, line 371: Can the text values be local to each method instead?
The reason is that several of the methods call each other and I'm worried you might at some point get nested usage of text.
It also seems better practice to keep them local instead of making the compiler worry about other methods maybe changing the text.

amtep (changes requested):
There's one merge conflict too :) maybe rebase

amtep:
The new stuff looks good to go

amtep (approved):

PR #92, opened 2024-03-23
Validate localization_values blocks

amtep:
This required some restructuring of the localization validator, because localization_values can be used as macros not just as scope values.


src/ck3/data/interactions.rs
```diff
@@ -78,6 +78,14 @@ impl DbKind for CharacterInteraction {
             }
         }
 
+        vd.field_validated_block("localization_values", |block, data| {
+            let mut vd = Validator::new(block, data);
+            vd.unknown_value_fields(|key, value| {
+                let scopes = validate_target(value, data, &mut sc, Scopes::all());
```
my4ng, line 84: `this` is possible as the RHS, so use `validate_target_ok_this` instead

src/validator.rs
```diff
@@ -347,6 +347,23 @@ impl<'a> Validator<'a> {
...
+    ///
+    /// Expect no more than one `name` field in the block.
+    /// Returns true iff the field is present.
+    #[allow(dead_code)]
+    pub fn field_localization(&mut self, name: &str, sc: &mut ScopeContext) -> bool {
+        let sev = self.max_severity;
+        self.field_check(name, |_, bv| {
+            if let Some(token) = bv.expect_value() {
+                self.data.verify_exists_max_sev(Item::Localization, token, sev);
+                self.data.localization.validate_use(token.as_str(), self.data, sc);
```
my4ng, line 362: Will this give double error if the loca key does not exist?

src/trigger.rs
```diff
@@ -1106,16 +1112,22 @@ pub fn validate_target_ok_this(
...
+    token: &Token,
+    data: &Everything,
+    sc: &mut ScopeContext,
+    outscopes: Scopes,
+) -> Scopes {
     if token.is("this") {
         let msg = "target `this` makes no sense here";
         warn(ErrorKey::UseOfThis).msg(msg).loc(token).push();
     }
+    validate_target_ok_this(token, data, sc, outscopes)
```
my4ng, line 1130: I think we can skip this in the else branch

src/datatype.rs
```diff
@@ -203,10 +203,7 @@ pub struct Code {
 }
 
 /// `CodeArg` represents a single argument of a [`Code`].
-// Possibly the literal arguments can themselves contain [ ] code blocks.
-// I'll have to test that.
-// A literal argument can be a string that starts with a (datatype) in front
-// of it, such as '(int32)0'.
+// TODO: Verify whether the literal arguments can themselves contain [ ] code blocks.
```
my4ng, line 206: I think we can say with some degree of caution that it can't

src/data/localization.rs
```diff
@@ -173,19 +120,33 @@ impl LocaEntry {
                 match macrovalue {
                     MacroValue::Text(ref token) => vec.push(token.clone().linked(link)),
                     MacroValue::Keyword(k, _) => {
```
my4ng, line 122: maybe replace this with `keyword` or `key`

src/data/localization.rs
```diff
@@ -412,7 +376,7 @@ impl Localization {
             LocaValue::ComplexTooltip(tag, token, _) => {
                 // TODO: if any of the three are datatype expressions, validate them.
                 #[cfg(feature = "ck3")]
-                if Game::is_ck3() && !token.starts_with("[") && !token.starts_with("$") {
+                if Game::is_ck3() && !token.starts_with("[") && !is_builtin_macro(token) {
```
my4ng, line 379: Would it be better keep these in `tables/localization`?

src/data/localization.rs
```diff
@@ -492,7 +456,9 @@ impl Localization {
                 // TODO: - imperator -
```
my4ng, line 456: Should we add imperator support here?

src/data/localization.rs
```diff
@@ -492,7 +456,9 @@ impl Localization {
...
             }
             LocaValue::Icon(token) => {
-                data.verify_exists(Item::TextIcon, token);
+                if !is_builtin_macro(token) && !token.is("ICONKEY_icon") && !token.is("KEY_icon") {
```
my4ng, line 459: Are the last two allowed? I did not find any reference in vanilla games

src/data/localization.rs
```diff
@@ -537,29 +503,54 @@ impl Localization {
...
     pub fn validate_pass2(&self, data: &Everything) {
         scope(|s| {
-            // Hold the lock for the whole validation loop, to avoid the overhead of re-acquiring it
-            let already_validated = self.keys_validated_with_sc.read().unwrap();
             for (lang, hash) in &self.locas {
                 for entry in hash.values() {
-                    if !already_validated.contains(entry.key.as_str()) {
+                    if !entry.validated.load(Relaxed) {
                         s.spawn(|_| {
                             let mut sc = ScopeContext::new_unrooted(Scopes::all(), &entry.key);
```
my4ng, line 551: maybe a good idea to comment out `entry.validated.store(true, Relaxed)` but note it's not needed after this anyway?

my4ng (changes requested):
Just a few small problems and questions

src/ck3/data/interactions.rs
amtep: I think it's a case where `this` makes no sense because in character interactions there's no root

src/trigger.rs
amtep: I still want to check if `this` is the right scope :)

src/data/localization.rs
amtep: Look for $ICONKEY$_icon and $KEY$_icon

src/validator.rs
amtep: No, validate_use is silent if the key is not defined

src/datatype.rs
amtep: This change will be gone after rebase:)

src/data/localization.rs
amtep: done

src/data/localization.rs
amtep: Yeah but that should be a separate PR

src/data/localization.rs
amtep: Yeah I added a comment

src/data/localization.rs
amtep: Yeah, I'll make a table

src/ck3/data/interactions.rs
my4ng: Sorry you are correct. I was thinking about the example in health_effects

src/data/localization.rs
my4ng: Unfortunately still can't any; do you have a specific file example?

my4ng (approved):

PR #93, opened 2024-03-23
Self-update binaries

my4ng:


src/update.rs
```diff
@@ -0,0 +1,85 @@
...
+            version
+        } else {
+            let releases = ReleaseList::configure()
+                .repo_owner("amtep")
+                .repo_name("ck3-tiger")
+                .with_target(consts::OS)
+                .build()?
+                .fetch()?;
+
+            releases.first().ok_or(UpdateError::MissingRelease)?.version.clone()
```
amtep, line 55: I think this needs a few more smarts, to skip releases that are "draft" or "pre-release"

amtep (changes requested):

src/update.rs
my4ng: Right now this is a little cursed by the limitation of `self_update` we discussed before. We can filter the releases and get the first desirable release, and specify that in the following updater. However, in doing so, it would be a forced-update even if it is not needed.

my4ng: I've looked deeper into `self_update` and it seems that draft and prerelease booleans are currently not used. This again requires upstream changes to function properly.

PR #100, opened 2024-03-31
refactor: easier API for localization ValueParser

amtep:


src/parse/localization.rs
```diff
@@ -783,7 +783,7 @@ impl<'a> ValueParser<'a> {
         self.value.push(LocaValue::Text(text.take_to_token()));
     }
 
-    pub fn parse_value(&mut self) -> Vec<LocaValue> {
+    pub fn parse_vec(&mut self) -> Vec<LocaValue> {
```
my4ng, line 786: Can this and `parse` use `(self)` instead as to avoid take?

my4ng (dismissed):

my4ng (approved):

PR #107, opened 2024-04-08
Refactor tiger binaries into `tiger-bin-shared` 

my4ng:
This also separates the self-update functionality into the aforementioned package.


ck3-tiger/Cargo.toml
```diff
@@ -15,17 +15,15 @@ rust-version = "1.70"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-tiger-lib = { version = "0.9.5", path = "..", features = ["ck3"] }
+tiger-lib = { path = "..", features = ["ck3"] }
```
amtep, line 18: Hmm you removed the version. Will this still pull in the latest tiger-lib when updating ck3-tiger?

tiger-bin-shared/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+ck3 = ["tiger-lib/ck3"]
+vic3 = ["tiger-lib/vic3"]
+imperator = ["tiger-lib/imperator"]
+
+[target.'cfg(target_os = "windows")'.dependencies]
+self_update = { version = "0.39", features = [
+    "archive-zip",
+    "compression-zip-deflate",
+]}
+winreg = "0.52"
```
amtep, line 34: For some reason the CI check complains about lack of winreg

tiger-bin-shared/src/binary.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// installed over the current binary.
+    Update {
+        /// release version (e.g. 0.9.3)
+        version: Option<String>,
+    },
+}
+
+#[derive(Args)]
+struct ValidateArgs {
+    /// Path to .mod file of mod to check.
```
amtep, line 56: This is awkward because Vic3 has no .mod files

amtep (changes requested):
Can you move the gamedir module as well?

And can you do the _game_-tiger-auto binaries? Imperator doesn't currently have one but it should, for symmetry.

ck3-tiger/Cargo.toml
my4ng: yes you are right won't be able to publish on crates.io otherwise. will fix

amtep (dismissed):

amtep (approved):

PR #115, opened 2024-05-08
Update effect_validation.rs - Gold field is not required for create_inspiration effect.

mrtats:
Gold is not a required field for create_inspiration effect. It is used to override the base cost of the given inspiration type.


amtep (approved):
Thanks!

PR #126, opened 2024-09-02
feat(vic3): validate create_building

### LINKED ISSUE #125: Check whether the items of `activate_production_methods` inside `create_building` is valid for the given building type

amtep:
Also fixes #125

### DIFF
--- a/src/vic3/data/buildings.rs
+++ b/src/vic3/data/buildings.rs
@@ -12,6 +12,7 @@ use crate::token::Token;
 use crate::tooltipped::Tooltipped;
 use crate::trigger::validate_trigger;
 use crate::validator::Validator;
+use crate::vic3::data::production_methods::ProductionMethodGroup;
 
 #[derive(Clone, Debug)]
 pub struct BuildingType {}
@@ -23,6 +24,29 @@ inventory::submit! {
 impl BuildingType {
     pub fn add(db: &mut Db, key: Token, block: Block) {
         db.add(Item::BuildingType, key, block, Box::new(Self {}));
+    }
+
+    #[allow(clippy::unused_self)]
+    pub fn validate_production_method(
+        &self,
+        pm: &Token,
+        building: &Token,
+        block: &Block,
+        data: &Everything,
+    ) {
+        if let Some(groups) = block.get_field_list("production_method_groups") {
+            for group in groups {
+                if let Some((_, block, group_item)) = data
+                    .get_item::<ProductionMethodGroup>(Item::ProductionMethodGroup, group.as_str())
+                {
+                    if group_item.contains_production_method(pm, block, data) {
+                        return;
+                    }
+                }
+            }
+        }
+        let msg = format!("production method `{pm}` not valid for `{building}`");
+        err(ErrorKey::Validation).msg(msg).loc(pm).push();
     }
 }
 
--- a/src/vic3/data/production_methods.rs
+++ b/src/vic3/data/production_methods.rs
@@ -99,6 +99,19 @@ impl ProductionMethodGroup {
     pub fn add(db: &mut Db, key: Token, block: Block) {
         db.add(Item::ProductionMethodGroup, key, block, Box::new(Self {}));
     }
+
+    #[allow(clippy::unused_self)]
+    pub fn contains_production_method(
+        &self,
+        contains_pm: &Token,
+        block: &Block,
+        _data: &Everything,
+    ) -> bool {
+        if let Some(pms) = block.get_field_list("production_methods") {
+            return pms.iter().any(|pm| pm == contains_pm);
+        }
+        false
+    }
 }
 
 impl DbKind for ProductionMethodGroup {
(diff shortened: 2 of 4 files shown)


src/vic3/data/buildings.rs
```diff
@@ -24,6 +25,29 @@ impl BuildingType {
...
+        if let Some(groups) = block.get_field_list("production_method_groups") {
+            for group in groups {
+                if let Some((_, block, group_item)) = data
+                    .get_item::<ProductionMethodGroup>(Item::ProductionMethodGroup, group.as_str())
+                {
+                    if group_item.contains_production_method(pm, block, data) {
+                        return;
+                    }
+                }
+            }
```
my4ng, line 46: This looks like only the first group is checked if it does contain production method. Is this the desired behaviour?

src/vic3/data/production_methods.rs
```diff
@@ -99,6 +99,19 @@ impl ProductionMethodGroup {
...
+    pub fn contains_production_method(
+        &self,
+        contains_pm: &Token,
+        block: &Block,
+        _data: &Everything,
+    ) -> bool {
+        if let Some(pms) = block.get_field_list("production_methods") {
+            return pms.iter().any(|pm| pm == contains_pm);
+        }
+        false
```
my4ng, line 113: can be replaced by `is_some_and` but not a big deal

src/vic3/effect_validation.rs
```diff
@@ -243,6 +244,57 @@ pub fn validate_country_value(
...
+        if let Some(building) = building {
+            if let Some((_, block, building_item)) =
+                data.get_item::<BuildingType>(Item::BuildingType, building.as_str())
+            {
+                building_item.validate_production_method(token, building, block, data);
+            }
+        }
+    });
+    vd.field_bool("subsidized");
+    vd.field_numeric_range("reserves", 0.0..=1.0);
```
my4ng, line 269: There are no `reserves=0.*` in vanilla, but there is one with `=2`. Maybe this is in absolute unit?

src/vic3/effect_validation.rs
```diff
@@ -243,6 +244,57 @@ pub fn validate_country_value(
...
+    vd.field_validated_value("level", |_, mut vd| {
+        vd.maybe_is("arable_land");
```
my4ng, line 271: didn't find a reference to this usage in vanilla

src/vic3/effect_validation.rs
```diff
@@ -243,6 +244,57 @@ pub fn validate_country_value(
...
+        vd.integer();
+    });
+    vd.field_validated_block("add_ownership", |block, data| {
+        let mut vd = Validator::new(block, data);
+        vd.multi_field_validated_block("country", |block, data| {
+            let mut vd = Validator::new(block, data);
+            vd.req_field("country");
+            vd.field_target("country", sc, Scopes::Country);
+            vd.req_field("levels");
+            vd.field_script_value("levels", sc);
```
my4ng, line 281: might be hardcoded `field_integer`

src/vic3/effect_validation.rs
```diff
@@ -243,6 +244,57 @@ pub fn validate_country_value(
...
+        });
+        // Docs say "country" for both, but vanilla uses "building".
+        vd.multi_field_validated_block("building", |block, data| {
+            let mut vd = Validator::new(block, data);
+            vd.req_field("country");
+            vd.field_target("country", sc, Scopes::Country);
+            vd.req_field("levels");
+            vd.field_script_value("levels", sc);
```
my4ng, line 289: same

my4ng (changes requested):

src/vic3/data/buildings.rs
amtep: Hmm I don't see it. It's supposed to exit if any of the groups contain that production method.

src/vic3/effect_validation.rs
amtep: I got this from the effect documentation. I don't know why there's a reserves=2 or what it's supposed to mean, i figured it was a bug in vanilla.

src/vic3/effect_validation.rs
amtep: It was also in the effect documentation :)

src/vic3/effect_validation.rs
amtep: Yeah I don't know so I went with the less restrictive one. I guess I could test it

dragon-archer: I've tested this in my mod, seems that both `level` in `create_building` and `levels` in `add_ownership` inside `create_building` accept only `field_integer`, while `script value`, `value scope` and `variable` will all generate an error.

P.S. Actually this is very annoying, so I even create a `scripted effect` to simulate `create_building` while accepting any kinds of `level` by enuming the `level` and expand it to `create_building` call with `field_integer`

amtep: Many thanks for testing it :)

src/vic3/data/production_methods.rs
amtep: I think that would actually make the code harder to read, with its nested closures

PR #128, opened 2024-09-16
Vic3 event and trigger correction

HiddeLekanne:


src/vic3/data/events.rs
```diff
@@ -155,6 +155,7 @@ impl Event {
         let mut vd = Validator::new(&self.block, data);
 
         let mut tooltipped_immediate = Tooltipped::Past;
+        let mut tooltipped_after = Tooltipped::Past
```
amtep: Are you sure the `after` block is tooltipped as past? It will be run after the event completes, so it seems like it would use the normal tooltipped value.

amtep (changes requested):

PR #131, opened 2024-09-21
Column overflow fix

HiddeLekanne:
Some files Vic 3 files have big one-liners (like "common\terrain_manipulators\provinces\allowed_provinces.txt"). So you actually do need the extra bits on your column tracker.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea/
+.vscode/
 /target/
 */benches/config.toml
 cachegrind.out*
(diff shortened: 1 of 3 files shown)


amtep:
This is unfortunate because currently `Loc` is 16 bytes which makes it cheap to copy. Bumping it to 20 bytes will slow down the program. I've been thinking of dividing the bytes between `line` and `column` (maybe 3 bytes each) but that might slow down the program in other ways because it requires calculations when incrementing them.

amtep (approved):

PR #133, opened 2024-09-22
Diplomatic play and trigger fix

HiddeLekanne:
diplo play effects have the diplo play as scope
Also the radical fraction trigger exists

### DIFF
--- a/src/vic3/data/diplomatic_plays.rs
+++ b/src/vic3/data/diplomatic_plays.rs
@@ -31,6 +31,10 @@ impl DbKind for DiplomaticPlay {
         sc.define_name("target_country", Scopes::Country, key);
         sc.define_name("initiator", Scopes::Country, key); // undocumented
 
+        let mut diplo_sc = ScopeContext::new(Scopes::DiplomaticPlay, key);
+        diplo_sc.define_name("target_country", Scopes::Country, key);
+        diplo_sc.define_name("initiator", Scopes::Country, key); // undocumented
+
         data.verify_exists(Item::Localization, key);
         let loca = format!("{key}_desc");
         data.verify_exists_implied(Item::Localization, &loca, key);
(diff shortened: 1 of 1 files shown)


src/vic3/tables/triggers.rs
```diff
@@ -924,6 +924,7 @@ const TRIGGER: &[(Scopes, &str, Trigger)] = &[
     (Scopes::Pop, "pop_loyalist_fraction", CompareValue),
     (Scopes::Pop, "pop_neutral_fraction", CompareValue),
     (Scopes::Pop, "pop_radical_fraction", CompareValue),
+    (Scopes::Country.union(Scopes::State), "radical_fraction", CompareValue),
```
amtep, line 927: radical_fraction is already in the table, lower down

amtep (changes requested):
The first commit is good, the second doesn't look okay.

Can you make your commit messages follow the format of the others in the commit history? Especially starting the commit with fix(vic3) or feat(vic3) to indicate it's a vic3 commit.

src/vic3/tables/triggers.rs
HiddeLekanne: Aah right, did it in the completely wrong place. 

`	value = radical_fraction`
Is also valid code. I don't know where to add that.

amtep: There's no place to add that because that's new syntax :) Not supported yet.

The usual place would be the TRIGGER_COMPLEX table, but that expects `(...)` after the trigger name.

HiddeLekanne: The problem it's not a trigger, it's an actual value assignment.

amtep: It might make sense to just add it as a special case in `validate_trigger_key_bv` and `validate_target_ok_this` for now.

amtep: > The problem it's not a trigger, it's an actual value assignment.

Yeah that would be covered by treating it as a 'target', and triggers with `CompareValue` outputs can be used as targets.

`radical_fraction` is unique in being able to _optionally_ take a `Block`.

amtep: One solution might be to add a `BlockOrCompareValue` variant to the `Trigger` enum, which checks whether it's getting a `Block` or a `Value`. And update the `trigger_comparevalue` function to treat it as a `CompareValue`.

HiddeLekanne:
Did both commit rename and removed the thing I still don't know how to fix

PR #136, opened 2024-10-01
fix(vic3) building validation scopes

HiddeLekanne:
In addition to this fix, `ai_nationalization_desire `seems to be a value of some sort, but I am not sure how to add that.


amtep:
> In addition to this fix, `ai_nationalization_desire `seems to be a value of some sort, but I am not sure how to add that.

Do you mean it can't be a script value? In that case field_numeric is the way to go.

amtep (approved):

HiddeLekanne:
More like I didn't know what to do with the advice_field, if it were to convert to field_numeric.

PR #141, opened 2024-10-07
Implement checks for cultures and split_religion entries in create_pop

Lunar-Dawn:
This should allow for and verify usage of the proportional syntax when creating pops, I've found it very useful when populating lots of states when I only care about the total population and vague demographics.

There's still the question mark about whether one should check that the numbers add up to ~1, if that's worth a warning. I'd think so, but I worry about spam.

I'm also a bit uncertain about best practices when emitting errors, so they might need correcting.

### DIFF
--- a/src/validator.rs
+++ b/src/validator.rs
@@ -1,4 +1,5 @@
 use std::borrow::Borrow;
+use std::collections::HashMap;
 use std::fmt::{Debug, Display, Formatter};
 use std::ops::{Bound, RangeBounds};
 use std::str::FromStr;
@@ -1544,6 +1545,56 @@ impl<'a> Validator<'a> {
                 if let Some(block) = bv.expect_block() {
                     f(date, key, block, self.data);
                 }
+            }
+        }
+    }
+
+    /// Expect the block to contain any number of `key = value` fields
+    /// where each key is a unique Item of type itype.
+    /// Run the closure `f(key, vd)` for every matching block.
+    #[cfg(feature = "vic3")]
+    pub fn validate_item_key_values<F>(&mut self, itype: Item, mut f: F)
+    where
+        F: FnMut(&Token, ValueValidator),
+    {
+        let mut visited_fields = HashMap::new();
+        for Field(key, _, bv) in self.block.iter_fields() {
+            self.data.verify_exists(itype, key);
+
+            match visited_fields.get(key.as_str()) {
+                Some(&duplicate) => dup_assign_error(key, duplicate),
+                None => { visited_fields.insert(key.as_str(), key); },
+            }
+
+            self.known_fields.push(key.as_str());
+            if let Some(value) = bv.expect_value() {
+                let mut vd = ValueValidator::new(value, self.data);
+                vd.set_max_severity(self.max_severity);
+                f(key, vd);
+            }
+        }
+    }
+
+    /// Expect the block to contain any number of `key = { block }` fields
+    /// where each key is a unique Item of type itype.
+    /// Run the closure `f(key, block, data)` for every matching block.
+    #[cfg(feature = "vic3")]
+    pub fn validate_item_key_blocks<F>(&mut self, itype: Item, mut f: F)
+    where
+        F: FnMut(&Token, &Block, &Everything),
+    {
+        let mut visited_fields = HashMap::new();
+        for Field(key, _, bv) in self.block.iter_fields() {
+            self.data.verify_exists(itype, key);
+
+            match visited_fields.get(key.as_str()) {
+                Some(&duplicate) => dup_assign_error(key, duplicate),
+                None => { visited_fields.insert(key.as_str(), key); },
+            }
+
+            self.known_fields.push(key.as_str());
+            if let Some(block) = bv.expect_block() {
+                f(key, block, self.data);
             }
         }
     }
--- a/src/validator.rs
+++ b/src/validator.rs
@@ -1588,7 +1588,7 @@ impl<'a> Validator<'a> {
                 vd.set_max_severity(sev);
                 f(key, vd);
             }
-        })
+        });
     }
 
     /// Expect the block to contain any number of `key = { block }` fields
@@ -1603,7 +1603,7 @@ impl<'a> Validator<'a> {
             if let Some(block) = bv.expect_block() {
                 f(key, block, data);
             }
-        })
+        });
     }
 
     /// Expect the block to contain any number of unknown fields (so don't warn about unknown fields anymore).
--- a/src/validator.rs
+++ b/src/validator.rs
@@ -1548,6 +1548,31 @@ impl<'a> Validator<'a> {
         }
     }
 
+    /// Expect the block to contain any number of `key = value` or `key = { block }`fields
+    /// where each key is a unique Item of type itype.
+    /// Run the closure `f(key, bv, data)` for every matching block.
+    #[allow(dead_code)]
+    fn validate_item_key_fields<F>(&mut self, itype: Item, mut f: F)
+    where
+        F: FnMut(&Token, &BV, &Everything),
+    {
+        let mut visited_fields = TigerHashSet::default();
+        for Field(key, _, bv) in self.block.iter_fields() {
+            self.data.verify_exists(itype, key);
+
+            match visited_fields.get(key.as_str()) {
+                Some(&duplicate) => dup_assign_error(key, duplicate),
+                None => {
+                    visited_fields.insert(key);
+                }
+            }
+
+            self.known_fields.push(key.as_str());
+
+            f(key, bv, self.data);
+        }
+    }
+
     /// Expect the block to contain any number of `key = value` fields
     /// where each key is a unique Item of type itype.
     /// Run the closure `f(key, vd)` for every matching block.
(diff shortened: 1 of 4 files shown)


amtep:
Hmm this is the first use case for modeling pdx's CFixedPoint type :) It would allow an exact calculation to see if the fractions add up to 1. Currently, tiger approximates the numbers with an f64.

src/validator.rs
```diff
@@ -1548,6 +1549,56 @@ impl<'a> Validator<'a> {
...
 
+    /// Expect the block to contain any number of `key = value` fields
+    /// where each key is a unique Item of type itype.
+    /// Run the closure `f(key, vd)` for every matching block.
+    #[cfg(feature = "vic3")]
+    pub fn validate_item_key_values<F>(&mut self, itype: Item, mut f: F)
+    where
+        F: FnMut(&Token, ValueValidator),
+    {
+        let mut visited_fields = HashMap::new();
```
amtep, line 1560: This should use TigerHashMap (from crate::helpers) for consistency with other hashmaps.

src/validator.rs
```diff
@@ -1548,6 +1549,56 @@ impl<'a> Validator<'a> {
...
+
+    /// Expect the block to contain any number of `key = { block }` fields
+    /// where each key is a unique Item of type itype.
+    /// Run the closure `f(key, block, data)` for every matching block.
+    #[cfg(feature = "vic3")]
+    pub fn validate_item_key_blocks<F>(&mut self, itype: Item, mut f: F)
+    where
+        F: FnMut(&Token, &Block, &Everything),
+    {
+        let mut visited_fields = HashMap::new();
```
amtep, line 1586: Same

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+    // TODO: Consider doing sanity checks on the numbers in the splits
+    //  Since they're percentages of the size, not ratios, if they don't add up to 1 you'll get unexpected results
+    //  But that would involve float maths and comparisons for something that isn't technically wrong.
+
+    let mut available_cultures = HashSet::new();
+
+    if block.has_key("culture") {
+        vd.ban_field("cultures", || "pops with several cultures");
+        vd.field_item("culture", Item::Culture);
+        if let Some(token) = vd.field_value("culture") {
```
amtep, line 579: Would be simpler to get the token by using `if let Some(token) = block.get_field_value("culture")` above instead of just `has_key`

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+            available_cultures.insert(token.as_str());
+        }
+    } else {
+        vd.field_validated_block("cultures", |block, data| {
+            let mut vd = Validator::new(block, data);
+            vd.validate_item_key_values(Item::Culture, |key, mut vd| {
+                available_cultures.insert(key.as_str());
+                vd.numeric()
```
amtep, line 587: `numeric_range(0.0..=1.0)` would be nice here. It doesn't exist yet in ValueValidator, but Validator has the corresponding field function

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+                        let msg = format!("Trying to split religion of culture {} multiple times", key.as_str());
+                        let msg_other = "First split here";
+                        err(ErrorKey::DuplicateField).msg(msg).loc(key).loc_msg(duplicate, msg_other).push();
+                    },
+                    None => { used_cultures.insert(key.as_str(), key.clone()); },
+                }
+
+                let mut vd = Validator::new(block, data);
+                vd.validate_item_key_values(Item::Religion, |_, mut vd| {
+                    vd.numeric();
```
amtep, line 625: Same about the range

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+                vd.numeric()
+            })
+        });
+    }
+
+    if block.has_key("religion") {
+        vd.ban_field("split_religion", || "pops without a `religion` field");
+        vd.field_item("religion", Item::Religion);
+    } else if block.has_key("split_religion") {
+        let mut used_cultures = HashMap::new();
```
amtep, line 596: Same about the HashMap

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+                only_one_culture = true;
+
+                if !available_cultures.contains(key.as_str()) {
+                    let msg = "culture being split does not appear in pop";
+                    err(ErrorKey::FieldMissing).msg(msg).loc(key).push();
+                }
+
+                match used_cultures.get(key.as_str()) {
+                    Some(duplicate) => {
+                        let msg = format!("Trying to split religion of culture {} multiple times", key.as_str());
```
amtep, line 616: Tiger generally starts error messages with lowercase (though it's not completely consistent on this)

amtep: Also, you can just do `{key}` because token has a Display impl

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+                        let msg_other = "First split here";
```
amtep, line 617: Should also be lowercase

Lunar-Dawn:
It would be fairly trivial to parse as well, simply parse the two sides of the decimal as integers. Would it be worth to make it into a completely new type to model it, or check if an i64 equals 100'000?

amtep (changes requested):
Looks very good :) I left some comments though

src/validator.rs
```diff
@@ -1548,6 +1549,56 @@ impl<'a> Validator<'a> {
         }
     }
 
+    /// Expect the block to contain any number of `key = value` fields
+    /// where each key is a unique Item of type itype.
+    /// Run the closure `f(key, vd)` for every matching block.
+    #[cfg(feature = "vic3")]
+    pub fn validate_item_key_values<F>(&mut self, itype: Item, mut f: F)
```
my4ng, line 1556: Can group the common parts from the two into a base function

src/vic3/effect_validation.rs
Lunar-Dawn: I was under the incorrect assumption that would complain if `culture` didn't exist. I'll do that then, the old solution felt a bit dirty.

amtep:
> Yeah, it would be fairly trivial to parse as well, simply parse the two sides of the decimal as integers. Would it be worth to make it into a completely new type to model it, or check if an i64 equals 100'000?

I'd prefer it as its own type, because eventually we'll want to use it in the parser for the `@[...]` calculations. Which means supporting multiply and divide. The last time I looked into this, there were some crates for fixed point.

It's okay to do something simpler for now and leave it as a TODO, though. Tiger is an endless TODO :)

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+                        err(ErrorKey::DuplicateField).msg(msg).loc(key).loc_msg(duplicate, msg_other).push();
+                    },
+                    None => { used_cultures.insert(key.as_str(), key.clone()); },
```
my4ng, line 620: From memory, `Token` implements `Borrow<str>` and hence `used_cultures` can have type `TigerHashSet<Token>`.

src/vic3/effect_validation.rs
```diff
@@ -555,18 +556,77 @@ pub fn validate_create_military_formation(
...
+        vd.multi_field_validated_block("split_religion", |block, data| {
+            let mut vd = Validator::new(block, data);
+            let mut only_one_culture = false;
+
+            vd.validate_item_key_blocks(Item::Culture, |key, block, data| {
+                if only_one_culture {
+                    let msg = "split_religion should contain only one culture block";
+                    err(ErrorKey::DuplicateItem).msg(msg).loc(key).push();
+                }
+                only_one_culture = true;
```
my4ng, line 607: Doesn't seem to have a check for when there is *no* culture outside

src/vic3/effect_validation.rs
Lunar-Dawn: I had no idea Rust's `HashSet`s could do that, never touched the language before. That would be a fair bit of boilerplate in C++

src/validator.rs
```diff
@@ -1548,6 +1549,56 @@ impl<'a> Validator<'a> {
         }
     }
 
+    /// Expect the block to contain any number of `key = value` fields
+    /// where each key is a unique Item of type itype.
+    /// Run the closure `f(key, vd)` for every matching block.
+    #[cfg(feature = "vic3")]
```
amtep, line 1554: It's okay to do `#[allow(dead_code)]` instead. That way you don't get the warning about unused import of TigerHashSet. (It's because it is unused when not compiling for vic3)

amtep:
You can do `cargo fmt` to automatically format the code. That gets rid of the complaints from the rustfmt check :)

Lunar-Dawn:
I'll be honest, I hadn't even started looking at the CI errors. Fortunately, there weren't that many

amtep (dismissed):

amtep:
Now that the other checks passed, clippy got a chance :) It had some minor suggestions.

You can run it yourself with `cargo clippy -p vic3-tiger`

The fact that `cargo fmt` sometimes makes changes that then make clippy complain is a known annoyance, but it's easily solved by adding the semicolon clippy asks for.

If you want, you can install the sample pre-commit hook that's in `git-hook-pre-commit` in the repo. That will run most of the CI checks locally when you commit.

Lunar-Dawn:
I think, *think* this is done now. Thank you for your immense patience while I've been flailing around with Rust. Turns out my "quick feature to code with Rust for the first time" was anything but.

The effect validator is immense now, and I'm unsure if it's because my code is bad, or if the effect really just is that complex.

src/vic3/effect_validation.rs
```diff
@@ -564,13 +564,20 @@ pub fn validate_create_pop(
 ) {
     // This effect is undocumented
 
+    fn sum_fractions_warner<E: ErrorLoc>(sum_fractions: i64, loc: E) {
+        if sum_fractions != 100_000 {
+            let msg = format!(
+                "fractions should add to exactly 1, currently {}.{}",
```
amtep, line 571: You should zero-pad the fraction here, otherwise 0.5 and 0.05 will both print as 0.5.

amtep (changes requested):
Found one bug :) Otherwise ok

Lunar-Dawn:
At least it was a very small fix, three characters

amtep (approved):

PR #150, opened 2024-10-20
fix(ck3): reroot ai_potential

### LINKED ISSUE #151: character_interactions/ai_potential lacks scopes
378152111-29545ff7-3d89-4b56-8cca-91cc256e4a9f

unLomTrois:
because otherwise it lacks scope context and doesn't see scopes like "target", "landed_title", as shown below:

image

The same vanilla code after fix:
image

I am not sure what you originally meant when you added a comment:
```
// This is in character scope with no other named scopes builtin
```

Closes #151

### DIFF
--- a/src/ck3/data/interactions.rs
+++ b/src/ck3/data/interactions.rs
@@ -298,9 +298,15 @@ impl DbKind for CharacterInteraction {
         vd.field_integer("ai_frequency"); // months
 
         // This is in character scope with no other named scopes builtin
-        vd.field_validated_block_rooted("ai_potential", Scopes::Character, |block, data, sc| {
-            validate_trigger(block, data, sc, Tooltipped::Yes);
-        });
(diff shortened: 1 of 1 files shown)


amtep:
The comment is based on the `.info` file:

> ai_potential = trigger                                                          # Will the ai consider trying this interaction.
>                                                                                                               # Root scope is the actor, and other event
>                                                                                                               # target are not available

"event target" is an older term for scopes.

The logic is that `ai_potential` is checked before all the work of selecting a recipient and a target is done, unlike `ai_will_do` which is evaluated afterward.

For additional evidence that the comment is still correct, see the `destroy_artifact_interaction`. It has a block in the `ai_will_do` with the comment:

> # This errors in the ai_potential so it's been moved here

indicating that `scope:target` is indeed not available there.

PR #163, opened 2025-02-01
fix(vic3) - Minor validation issues

kaiser-chris:
Hi @amtep,

I have fixed a few minor validation issues.
The only issue I have run into is that `onctrl` is still detected as an unknown GUI property. But the `wrap_count` worked so I am confused what I did wrong.

### Missing GUI properties `onctrl` and `wrap_count`
Added both properties and defined their content.
### Scope and Item for `has_technology_researched`
This trigger accepts both scopes and items even if the base game does not do it.
### Scope and Item for `create_character` religion and culture
Both the `religion` and `culture` fields accept items and scopes.
Verified it working with non primary cultures.
### Event Icons
The following event fields all accept a texture, a country as well as a character depending on the underlying event GUI window implementation:
 - `minor_left_icon`
 - `minor_right_icon`
 - `left_icon`
 - `right_icon`
 - `center_icon` (was missing)

Regards,

Chris

### DIFF
--- a/src/imperator/tables/modifs.rs
+++ b/src/imperator/tables/modifs.rs
@@ -50,7 +50,7 @@ pub fn lookup_modif(name: &Token, data: &Everything, warn: Option<Severity>) -> 
     // culture_$PopType$_happiness
     if let Some(part) = name_lc.strip_prefix_unchecked("culture_") {
         for &sfx in &["_output", "_happiness"] {
-            if let Some(part) = part.strip_prefix_unchecked(sfx) {
+            if let Some(part) = part.strip_suffix_unchecked(sfx) {
                 maybe_warn(Item::PopType, &part, name, data, warn);
                 return Some(ModifKinds::Country);
             }


src/vic3/effect_validation.rs
```diff
@@ -346,13 +346,8 @@ pub fn validate_create_character(
...
         warn(ErrorKey::Validation).msg(msg).loc(key).push();
     }
-    // NOTE: docs say this is an Item, but vanilla files consistently pass a scope.
-    vd.field_validated_value("culture", |_, mut vd| {
-        vd.maybe_is("primary_culture");
-        vd.target(sc, Scopes::Culture);
-    });
-    // TODO: vanilla files pass religion as an item in several places. Figure out if that's a bug.
-    vd.field_target("religion", sc, Scopes::Religion);
+    vd.field_item_or_target("culture", sc, Item::Culture, Scopes::Culture);
```
amtep, line 349: I think you still need the exception for a literal "primary_culture" here

kaiser-chris: Ah yes you are right.
I will add it.

kaiser-chris:
@amtep do you know what I did wrong with the definition of `onctrl`?

src/vic3/effect_validation.rs
kaiser-chris: Done

amtep: You fixed the wrong one :) You changed the imperator one

amtep:
> @amtep do you know what I did wrong with the definition of `onctrl`?

Nothing I noticed. I'll have to test.

amtep:
The clippy errors are due to github using a new Rust version. I'll fix them in main.

kaiser-chris:
> The clippy errors are due to github using a new Rust version. I'll fix them in main.

Ok the fmt errors and the imperator change should be fixed as well.

kaiser-chris:
Ohh rustfmt is more strict than expected. I have run it locally now before pushing.

amtep:
Yeah rustfmt doesn't produce ideal results but I like having the formatting be automated.  It's a tradeoff that I think produces more consistency in the long run, even though its line by line decisions can be inconsistent.

kaiser-chris:
> Yeah rustfmt doesn't produce ideal results but I like having the formatting be automated. It's a tradeoff that I think produces more consistency in the long run, even though its line by line decisions can be inconsistent.

Makes sense. I just haven't had any experience with rust so its strictness was kinda unexpected.

amtep:
`onctrl` was alphabetized wrong, breaking the binary search :) It should be after `oncreate`

PR #177, opened 2025-03-08
Localization language improvement

my4ng:
~10% performance uplift


src/data/localization.rs
```diff
@@ -243,7 +243,7 @@ pub enum LocaValue {
     // The value is not stored in the enum because we don't validate it.
     // TODO: instead of Token here, maybe need Box<LocaValue> or a Vec<LocaValue>, or maybe a type
```
kaiser-chris, line 244: Is this TODO still relevant?

amtep: Yeah the idea of the TODO is to support tooltips where the tag is itself an expression with multiple parts

amtep (approved):

PR #208, opened 2025-07-04
fix(vic3): create_building can accept a script value for the level parameter

BrknRobot:
Docs still say level takes arable_land/integer. The 1.9.0 change notes say it can also take variables, but I've confirmed myself that math blocks work as well.


amtep (dismissed):

amtep:
Thanks!
Clippy says the code can be simplified, though :)

PR #210, opened 2025-07-06
Accept glob patterns in file filters

### LINKED ISSUE #209: Support Glob patters in file filters
It would be handy if I could use glob patterns in file filters. Particularly for localizations which frequently have similarly named files in many folders.

BrknRobot:
Closes #209

Existing file and folder matching behaviour is preserved.
Existing file filters which contain any of '?', '*', and '[' will require an update to escape these characters.
Most glob patterns will have to be quoted in the config, or the lexer will split them into different tokens.

### DIFF
--- a/src/report/filter.rs
+++ b/src/report/filter.rs
@@ -1,10 +1,12 @@
 use std::path::PathBuf;
+use glob::Pattern;
 
 use crate::block::Comparator;
 
 use crate::fileset::FileKind;
-use crate::report::{Confidence, ErrorKey, LogReport, Severity};
+use crate::report::{err, Confidence, ErrorKey, ErrorLoc, LogReport, Severity};
 use crate::token::Loc;
+use crate::Token;
 
 /// Determines whether a given Report should be printed.
 /// If a report is matched by both the blacklist and the whitelist, it will not be printed.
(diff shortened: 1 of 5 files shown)


src/report/filter.rs
```diff
@@ -112,12 +114,31 @@ impl FilterRule {
                 Comparator::AtMost => report.confidence <= *level,
             },
             FilterRule::Key(key) => report.key == *key,
-            FilterRule::File(path) => {
-                report.pointers.iter().any(|pointer| pointer.loc.pathname().starts_with(path))
+            FilterRule::File(pattern) => {
+                report.pointers.iter().any(|pointer|
+                    pattern.matches_path(pointer.loc.pathname())
+                    || pointer.loc.pathname().starts_with(PathBuf::from(pattern.as_str())))
```
amtep: Looks good :) There's just one issue here. The `PathBuf::from` call will create a new `PathBuf`, effectively copying the string, for every loop iteration.

You can instead use `Path::new` which is a cost free conversion that just wraps the existing string.

amtep (changes requested):

amtep:
Can you also update `filter.md` to explain the new behavior to users?

amtep (approved):

PR #211, opened 2025-07-07
Add support for loading secondary mods from steam's workshop directory

BrknRobot:
Adds a new field `workshop_id`/`workshop_modfile` to `load_mod` which contains the mods workshop ID.
This will load the secondary mod files from Steam's downloaded copy.
If both `mod`/`modfile` and `workshop_id`/`workshop_modfile` are defined, `mod`/`modfile` will be used falling back to `workshop_id`/`workshop_modfile` if the path can't be found.

Tested with Vic3 on Linux. The code path for other games may or may not be currently working

### DIFF
--- a/src/fileset.rs
+++ b/src/fileset.rs
@@ -318,7 +318,7 @@ impl Fileset {
                     add_loaded_mod_root(label);
                     self.loaded_mods.push(loaded_mod);
                 } else {
-                    bail!("could not load secondary mod from config; missing valid `modfile` or `workshop_modfile` field");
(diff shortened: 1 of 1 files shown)


amtep:
I'm getting an error using it with ck3; I'm debugging it now.
```
Error: could not load secondary mod from config; missing valid `modfile` or `workshop_modfile` field
```

amtep:
Oh the error was that I had configured `workship_id` :)

amtep (approved):

PR #224, opened 2025-07-10
fix(vic3): validate message popup_name

BrknRobot:


src/vic3/data/messages.rs
```diff
@@ -43,11 +43,7 @@ impl DbKind for Message {
 
         vd.field_integer("days");
 
-        if block.field_value_is("notification_type", "popup") {
-            vd.field_value("popup_name"); // TODO: find out what is allowed here
-        } else {
-            vd.ban_field("popup_name", || "`notification_type = popup`");
-        }
+        vd.field_item("popup_name", Item::WidgetName); // Maps to notification_popup
```
amtep, line 46: 1. I don't understand the comment :)
2. This change makes it accept `popup_name` even in messages that don't have `notification_type = popup`. Is that correct behavior?
3. Unfortunately `WidgetName` is currently broken. It finds widgets that have a `name` field directly in the top level widget, but not ones where the name field is further down or comes from a type definition etc.

Btw I would like to add you to `CONTRIBUTORS.md`. Under what name can I credit you?

amtep (changes requested):

amtep (approved):

PR #226, opened 2025-07-11
Suppress unnecessarily repeated errors

BrknRobot:
Some errors don't need to be reported more than once. For example, a missing item report is probably due either to single typo, or an actual missing item that's resolved by adding the item. In the later case, all instances of the error are resolved not at the error location, but by adding the item. Pointing out additional locations is unnecessary and noisy.

This adds the ability to suppress repeated errors, consolidating them into an "And {} other locations" line in the output.

I just had the thought that maybe it would have better to leave error collection unchanged, and only consolidate when building the output, but I've already done it this way, so here's a draft PR and you can tell me the approach is no good if you'd prefer something else.

```
warning(missing-item): custom localization random_free_state_of_place not defined in common/customizable_localization/
  --> [Vic3] localization/english/acw_text_l_english.yml
39 |  acw_events.7.t:0 "The Free State of [SCOPE.GetRootScope.GetCountry.GetCustom('random_free_state_of_place')]"
   |                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^
  --> And 21 other locations
  ```

### DIFF
--- a/src/report/errors.rs
+++ b/src/report/errors.rs
@@ -2,10 +2,10 @@
 
 use std::borrow::Cow;
 use std::cell::RefCell;
-use std::cmp::Ordering;
+use std::cmp::{min_by, Ordering};
 use std::fs::read;
 use std::io::Write;
-use std::iter::once;
+use std::iter::{empty, once};
 use std::mem::take;
 use std::ops::{Bound, RangeBounds};
 use std::path::PathBuf;
@@ -25,7 +25,19 @@ use crate::report::{
     ErrorKey, FilterRule, LogReport, LogReportMetadata, LogReportPointers, LogReportStyle,
     OutputStyle, PointedMessage,
 };
+use crate::set;
 use crate::token::{leak, Loc};
+
+/// Error types that should be logged once when consolidating reports
+static LOG_ONCE: LazyLock<TigerHashSet<ErrorKey>> = LazyLock::new(|| {
+    set!([
+        ErrorKey::MissingFile,
+        ErrorKey::MissingItem,
+        ErrorKey::MissingLocalization,
+        ErrorKey::MissingPerspective,
+        ErrorKey::MissingSound,
+    ])
+});
 
 static ERRORS: LazyLock<Mutex<Errors>> = LazyLock::new(|| Mutex::new(Errors::default()));
 
@@ -42,6 +54,7 @@ pub struct Errors<'a> {
 
     /// Determines whether a report should be printed.
     pub(crate) filter: ReportFilter,
+
     /// Output color and style configuration.
     pub(crate) styles: OutputStyle,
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/report/errors.rs [hunk 1]
```diff
@@ -39,6 +39,8 @@ pub struct Errors {
 
     /// Determines whether a report should be printed.
     pub(crate) filter: ReportFilter,
+    /// Determines whether an error type should be reported multiple times
+    pub(crate) log_once: Vec<ErrorKey>,
```
amtep, line 43: Should logically be a TigerHashSet, but a Vec may be faster in this case.

src/report/errors.rs
```diff
@@ -106,13 +109,28 @@ impl Errors {
         false
     }
 
+    fn should_log_once(&mut self, report: &LogReport) -> bool {
+        if self.log_once.contains(&report.key) {
+            for (stored_report, count) in &mut self.storage {
+                if stored_report.eq_locless(report) {
+                    *count += 1;
+                    return true;
+                }
```
amtep, line 118: I'm worried about the cost of this. Scanning the whole log again for every new entry makes building up the error log quadratic in time. Some mods do have tens of thousands of reports.

src/report/errors.rs
```diff
@@ -419,3 +437,14 @@ pub fn set_show_loaded_mods(v: bool) {
...
 }
+
+pub fn set_log_once_defaults() {
+    Errors::get_mut().log_once = vec![
+        // All instances of something missing are solved by adding the thing
+        ErrorKey::MissingFile,
+        ErrorKey::MissingItem,
+        ErrorKey::MissingLocalization,
+        ErrorKey::MissingPerspective,
+        ErrorKey::MissingSound,
```
amtep, line 448: `MissingSound` is not like the others. Very few modders will modify the sound banks. Instead, if Pdx removes a sound, the modders have to find a replacement sound to use instead.

tests/test.rs
```diff
@@ -1,11 +1,12 @@
 use std::path::PathBuf;
 use std::sync::{LazyLock, Mutex};
+use std::usize;
```
amtep, line 3: Rust docs say this is deprecated :) `usize` is a primitive type.

src/report/errors.rs
BrknRobot: Yeah, iterating the hashmap isn't really making full use of it's hash capabilities. What if I removed `pointers` from the LogReport struct, and changed `Errors.storage` to hold `TigerHashMap<LogReport,TigerHashSet<Vec<PointedMessage>>>`. That should let us do this in O(1) time again.

Also, not sure how many cases of identical report messages with different severity or confidence levels there would be, but another potential option would be moving severity, confidence, and pointers into their own `LogReportOccurrence` struct to make `Errors.storage` hold `TigerHashMap<LogReport,TigerHashSet<LogReportOccurrence>>`. Then while emitting reports we could select the occurrence with highest confidence and severity for display, while consolidating the remainder.

tests/test.rs
BrknRobot: Oops, looks like something my ide snuck in there

amtep:
> What prompted me to work on this was developing a new mod feature starting from a baseline of having no warnings.

Ok, that makes sense.

I think I prefer the default to list all reports. Maybe `--consolidate` instead of `--short` ? It's a bit more typing but it's clearer about what it does.

src/report/errors.rs
amtep: > Would it? As far as I can tell all `LogReport`'s are instantiated through the builder. Should just be able to update `ReportBuilderStage3.build()` and `ReportBuilderStage3.push()` without touching any other code

I'm not sure what you have in mind, but if you say so :)

> 🤔 Would this require maintaining a second hashset? Or is there a way to cast `TigerHashSet<LogReport>` to `TigerHashSet<LogReportWrapper>` without copying?

It would mean having a `TigerHashMap<LogReportWrapper,TigerHashSet<LogReport>>`, similar to one of your proposals.

src/parse/localization.rs
```diff
@@ -47,14 +47,15 @@ impl LocaParser {
...
+                warn(ErrorKey::Encoding).strong().msg(msg).info(info).abbreviated(loc).push();
                 offset += '\u{feff}'.len_utf8();
                 loc.column += 1;
                 chars.next();
             }
         } else {
-            warn(ErrorKey::Encoding).msg("Expected UTF-8 BOM encoding").loc(loc).push();
+            warn(ErrorKey::Encoding).msg("Expected UTF-8 BOM encoding").abbreviated(loc).push();
         }
+        loc.line += 1;
```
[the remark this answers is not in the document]
BrknRobot: I wanted the encoding errors to be whole file reports to be consistent with other encoding error messages.
You're right, I missed the column though, I'll get that put back.

src/report/errors.rs
BrknRobot: Do you still want MissingSound removed?

src/report/writer.rs
```diff
@@ -94,6 +99,17 @@ fn log_line_info<O: Write + Send>(errors: &Errors, output: &mut O, indentation:
     _ = writeln!(output, "{}", ANSIStrings(line_info));
 }
 
+/// Log the additional number of this error that were found in other locations
+fn log_count<O: Write + Send>(errors: &Errors, output: &mut O, indentation: usize, count: usize) {
+    let line_count: &[ANSIString<'static>] = &[
+        errors.styles.style(Styled::Default).paint(format!("{:width$}", "", width = indentation)),
+        errors.styles.style(Styled::Location).paint("-->"),
+        errors.styles.style(Styled::Default).paint(" "),
+        errors.styles.style(Styled::Location).paint(format!("And {count} other locations")),
```
amtep, line 108: "And" should be lowercase to match tiger's style (and because it's a sentence fragment)

src/report/errors.rs
```diff
@@ -27,6 +28,18 @@ use crate::report::{
 };
 use crate::token::{leak, Loc};
 
+/// Error types that should be logged once when consolidating reports
+static LOG_ONCE: LazyLock<TigerHashSet<ErrorKey>> = LazyLock::new(|| {
+    AHashSet::from([
```
amtep, line 33: Using AHashSet here still bothers me. Is there some variation that will work, like `[A, B, C].into()` or `[A, B, C].iter().collect()`  ?

amtep (changes requested):

src/report/errors.rs
```diff
@@ -104,26 +118,50 @@ impl Errors<'_> {
...
+                let mut iterator =
                     occurrences.iter().filter(|pointers| !self.should_ignore(report, pointers));
                 match report.style {
                     LogReportStyle::Full => {
-                        Box::new(iterator.map(move |pointers| (report, Cow::Borrowed(pointers))))
+                        if consolidate && LOG_ONCE.contains(&report.key) {
+                            if let Some(initial) = iterator.next() {
+                                let (pointers, additional_count) = iterator.fold(
+                                    (initial, 0usize),
+                                    |(first_occurence, count), e| {
```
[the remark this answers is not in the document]
BrknRobot: Really testing my ability to spell here

amtep (approved):

PR #244, opened 2025-07-22
Don't require mutability for logitally immutable get_line()

BrknRobot:
Retrieval from the file cache is a logically immutable operation. This hides the implementation details requiring mutability from the caller.


amtep:
I prefer to avoid merge commits. Can you rebase instead?

amtep (approved):

PR #251, opened 2025-07-22
Refactor output into param

BrknRobot:
Also included: A fixup that depended on the reduced mutability requirements

### DIFF
--- a/src/report/errors.rs
+++ b/src/report/errors.rs
@@ -3,15 +3,14 @@
 use std::borrow::Cow;
 use std::cell::RefCell;
 use std::cmp::Ordering;
-use std::fs::{read, File};
-use std::io::{stdout, Write};
+use std::fs::read;
+use std::io::Write;
 use std::iter::once;
 use std::mem::take;
 use std::ops::{Bound, RangeBounds};
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::sync::{LazyLock, Mutex, MutexGuard};
 
-use anyhow::Result;
 use encoding_rs::{UTF_8, WINDOWS_1252};
 
 use crate::helpers::{TigerHashMap, TigerHashSet};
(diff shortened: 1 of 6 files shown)


tiger-bin-shared/src/auto.rs
```diff
@@ -137,7 +137,7 @@ fn validate_mod(
     let output_filename =
         format!("{name_short}-tiger-{}.log", modpath.file_name().unwrap().to_string_lossy());
     let output_file = &logdir.join(output_filename);
-    set_output_file(output_file)?;
+    let mut output = Box::new(File::create(output_file)?);
```
amtep, line 140: Why is this boxed?

src/report/errors.rs
```diff
@@ -31,9 +30,8 @@ use crate::token::{leak, Loc};
 static ERRORS: LazyLock<Mutex<Errors>> = LazyLock::new(|| Mutex::new(Errors::default()));
 
 #[allow(missing_debug_implementations)]
+#[derive(Default)]
 pub struct Errors<'a> {
-    pub(crate) output: RefCell<Box<dyn Write + Send>>,
```
amtep, line 35: What's the rationale for removing this?

src/report/errors.rs
```diff
@@ -104,11 +104,9 @@ impl<'a> Errors<'_> {
     }
 
     /// Extract the stored reports, sort them, and return them as a vector of ([`LogReport`],Vec<[`PointedMessage`]>).
```
amtep, line 106: Comment is no longer accurate :)
Maybe just say "return them as a vector"

src/report/errors.rs
BrknRobot: The `get_mut()` call in `log_report_json()` requires a mutable reference to `Errors`. Removing `output` from the error struct avoids requiring any mutable reference to `Errors` in the writers, and is feasible to remove now that the legacy reporting functions are gone and writing to the output is triggered from a single calling function. Additionally, the output stream itself is not related to the collection and processing of reports, and including it in the Errors struct violates the single responsibility principle.

amtep (approved):

PR #258, opened 2025-07-26
perf: Localization

BrknRobot:
**Avoid unnecessary localisation database access**
The expensive operation here is key hashing. This PR tries to reduce `HashMap::get()` calls to 1 per language for each entry point to the Localization module. It does this by introducing `mark_used_lang()` which can be used by code that is already iterating over the languages, and having `mark_used()` and `mark_used_lang()` double as existence checks, removing the need to call `exists()` or `exists_lang()` in many places.

~~**Replaces bitarray storage with `[bool; N]`**
The primary reader of these fields is `Localization::iter_lang_idx()` which can't make use of bitwise parallelism. Additionally, it returns an iterator, giving poor temporal locality to the array access and negating any cache benefits. Replacing the bit arrays with standard bool arrays reduces total Ir cost of vic3-tiger by over 4%.~~

### DIFF
--- a/src/data/localization.rs
+++ b/src/data/localization.rs
@@ -276,7 +276,11 @@ fn get_file_lang(filename: &OsStr) -> Option<Language> {
 impl Localization {
     // TODO: Remove `+ '_` in edition 2024
     fn iter_lang_idx(&self) -> impl Iterator<Item = usize> + '_ {
-        (0..Language::COUNT).filter(|i| self.mod_langs[*i])
+        self.iter_lang().map(Language::to_idx)
+    }
+
+    fn iter_lang(&self) -> impl Iterator<Item = Language> + '_ {
+        Language::iter().filter(|i| self.mod_langs[i.to_idx()])
     }
 
     pub fn exists(&self, key: &str) -> bool {
@@ -313,13 +317,7 @@ impl Localization {
         if key.is_empty() {
             return;
         }
-        self.mark_used(key);
-        let mut langs_missing = Vec::new();
-        for lang in self.iter_lang_idx() {
-            if !self.locas[lang].contains_key(key) {
-                langs_missing.push(Language::from_idx(lang).into());
-            }
-        }
+        let langs_missing = self.mark_used_return_missing(key);
         if !langs_missing.is_empty() {
             let msg = format!("missing {} localization key {key}", stringify_list(&langs_missing));
             // TODO: get confidence level from caller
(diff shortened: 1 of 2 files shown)


my4ng:
Thank you for the PR! The second part is definitely a **big** improvement since hashing dominates the run time of tiger as we have found out previously. However, I have some reservations about the first part.

> **Replaces bitarray storage with `[bool; N]`**
> The primary reader of these fields is `Localization::iter_lang_idx()` which can't make use of bitwise parallelism. Additionally, it returns an iterator, giving poor temporal locality to the array access and negating any cache benefits. Replacing the bit arrays with standard bool arrays reduces total Ir cost of vic3-tiger by over 4%.

The `BitArray` implementation reduces the memory from 11 bytes to 2 bytes, which means the entire `*_langs*` can fit into a single 64-bit register, enabling further optimisation/simplification on the HW. I don't agree with the iterator argument which suggests it introduces runtime overhead; any abstraction *should* have been optimised away by the rust compiler and the LLVM backend to bitwise operations. This also explains the higher Ir count, which is again not necessarily a bad thing since ALU bitwise operations are very cheap in terms of CPU cycles. A much better real-world comparison would be the criterion benchmarks found in the `*-tiger` folders.

The PR *as a whole* has negligible effects on the run time in the benchmarks ( -1%), but produces a significant reduction (-30%) with the first part reverted (see [patch](https://github.com/amtep/tiger/tree/perf)). I would suggest reverting it.

criterion benchmarks

src/data/localization.rs
```diff
@@ -700,10 +696,9 @@ impl FileHandler<(Language, Vec<LocaEntry>)> for Localization {
                 if check.iter().any(|t| t.is(lang_str))
                     || (check.is_empty() && skip.iter().all(|t| !t.is(lang_str)))
                 {
-                    langs.set(lang.to_idx(), true);
+                    self.check_langs[lang.to_idx()] = true;
```
amtep, line 714: Note that there's a bug here which could affect the benchmarks. `check_langs` is initialized to all true. Iff there's a languages block, then it should only contain the languages configured there. So the `= true` is a no-op here because you removed the initialization to zero in this function.

So you may have been measuring mods that due to this bug are scanning 11x the usual number of languages.

amtep:
For reference, you can run these benchmarks with
```
cargo bench -p ck3-tiger
cargo bench -p vic3-tiger
```
And drop a config file in xxx-tiger/benches/config.toml

Unfortunately you'll have to look at the code to see the config file options

amtep:
I noticed a related problem and filed https://github.com/amtep/tiger/issues/259 :)

src/data/localization.rs
```diff
@@ -397,34 +390,58 @@ impl Localization {
...
+        let mut exists = false;
+        for lang in self.iter_lang() {
+            exists |= self.mark_used_lang(key, lang);
+        }
+        exists
+    }
+
+    /// Marks a localization key as used for all languages.
+    /// Returns a [`Vec<&str>`] containing the languages for which the key does not exist.
+    fn mark_used_missing(&self, key: &str) -> Vec<&'static str> {
```
amtep, line 405: I think `mark_used_return_missing` would be a better name for this Frankenstein function :)

amtep (approved):

PR #269, opened 2025-08-08
Refactor file handling

BrknRobot:
Goals
- Centralize file read logic
- Create a single owner of file content
- Remove duplication in the path table
- Preserve tracking of overridden files to enable future checks against missing items to track which file overwrite is to blame

### DIFF
--- a/src/fileset.rs
+++ b/src/fileset.rs
@@ -4,16 +4,19 @@ use std::borrow::ToOwned;
 use std::cmp::Ordering;
 use std::ffi::OsStr;
 use std::fmt::{Display, Formatter};
+use std::hash::{Hash, Hasher};
 use std::path::{Path, PathBuf};
 use std::string::ToString;
-use std::sync::RwLock;
+use std::sync::{Arc, OnceLock, RwLock};
 
 use anyhow::{bail, Result};
+use itertools::Itertools;
 use rayon::prelude::*;
 use walkdir::WalkDir;
 
 use crate::block::Block;
 use crate::everything::{Everything, FilesError};
+use crate::filecontent::FileContent;
 use crate::game::Game;
 use crate::helpers::TigerHashSet;
 use crate::item::Item;
@@ -24,10 +27,12 @@ use crate::modfile::ModFile;
 use crate::parse::ParserMemory;
 use crate::pathtable::{PathTable, PathTableIndex};
 use crate::report::{
-    add_loaded_dlc_root, add_loaded_mod_root, err, fatal, report, ErrorKey, Severity,
+    add_loaded_dlc_root, add_loaded_mod_root, err, fatal, report, store_source_file, ErrorKey,
+    Severity,
 };
 use crate::token::Token;
 use crate::util::fix_slashes_for_target_platform;
+use crate::TigerHashMap;
 
 /// Note that ordering of these enum values matters.
 /// Files later in the order will override files of the same name before them,
@@ -59,7 +64,7 @@ impl FileKind {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Debug)]
 pub struct FileEntry {
     /// Pathname components below the mod directory or the vanilla game dir
     /// Must not be empty.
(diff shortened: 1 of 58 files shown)


amtep:
This PR gives a significant speedup (6.5s to 5.0s for ck3) and I can't figure out why.

amtep:
There's something wrong with the loca parsing, which is probably why it's faster. There are no reports about localization files in the output.

amtep:
I was concerned about possible slowdown due to all the locking around `Arc<FileEntry>`, which is why I tried to benchmark it :)

In one of my early designs before the PathTable, I tried to make pathname in Loc an Arc, and it resulted in significant slowdown.

Other than that, I'll have to read it a few more times.

src/fileset.rs
```diff
@@ -374,6 +446,30 @@ impl Fileset {
         false
     }
 
+    pub fn is_finalized(&self) -> bool {
+        !self.ordered_files.is_empty()
+    }
+
+    fn path_to_files_key(path: &Path) -> PathBuf {
+        path.canonicalize().unwrap_or(path.to_owned())
```
amtep, line 454: canonicalize may be a problem.

The rust docs say:
> On Windows, this converts the path to use [extended length path](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file) syntax, which allows your program to use longer path names, but means you can only join backslash-delimited paths to it, and it may be incompatible with other applications (if passed to the application on the command-line, or written to a file another application may read).

Another potential issue is that if the game dir has symlinks in it for some reason, canonicalize will resolve those symlinks, breaking the directory prefix logic tiger uses.

I think it's better to leave it out. Tiger is already careful about the files it scans (because of aforementioned prefix logic), and the names of things like the launcher file don't have a big chance of being accessed under different aliases.

BrknRobot: I really just wanted a normalized path for the map key, not to use anywhere else. But it turns out the call to canonicalize was actually introducing a significant performance cost, I think because rust checks to see if the file exists. So I've taken it out and just used the provided full path. It will be up to the caller not to provide multiple paths to the same file.

src/files/fileset.rs
```diff
@@ -204,333 +286,101 @@ impl LoadedMod {
...
-
-    /// Extra CK3 directory loaded before vanilla.
-    #[cfg(feature = "jomini")]
-    jomini_root: Option<PathBuf>,
+    #[allow(dead_code)]
+    pub(super) paths: FilePaths,
 
     /// The mod being analyzed.
-    the_mod: LoadedMod,
+    #[allow(dead_code)]
```
amtep, line 293: Why these three dead_code allows? (the_mod, loaded_dlcs, and db). Surely they're used for all games.

src/files/fileset.rs
```diff
@@ -204,333 +286,101 @@ impl LoadedMod {
...
+            handler.load_file(entry, parser).map(|loaded| (entry, loaded))
         });
         for (entry, loaded) in entries {
-            handler.handle_file(&entry, loaded);
+            handler.handle_file(entry, loaded);
         }
         handler.finalize();
     }
 
+    /// # Panics
```
amtep, line 359: No need for this. The unwrap here is part of rust's "lock poisoning" feature. If a thread panics while holding a mutex, the mutex is marked as tainted. Unwrapping it ensures that all other threads will also panic if they try to take that lock, in order to ensure they don't encounter inconsistent data.

So it's not a real panic, it's a panic that's part of unrolling another panic.

src/files/fileset.rs
```diff
@@ -204,333 +286,101 @@ impl LoadedMod {
...
+
+    pub fn exists(&self, key: &str) -> bool {
+        self.entry(key).is_some()
     }
 
     pub fn iter_keys(&self) -> impl Iterator<Item = &Token> {
         self.filename_tokens.iter()
     }
 
+    /// # Panics
```
amtep, line 383: Same as above

src/files/fileset.rs
```diff
@@ -598,6 +448,7 @@ impl Fileset {
         }
     }
 
+    /// # Panics
```
amtep, line 451: Same as above

src/files/fileset.rs
```diff
@@ -647,7 +498,7 @@ impl Fileset {
                     if joined.starts_with(valid) {
                         let msg = format!("file in unexpected directory {}", dirname.display());
                         let info = format!("did you mean common/{} ?", dirname.display());
-                        err(ErrorKey::Filename).msg(msg).info(info).loc(entry).push();
+                        err(ErrorKey::Filename).msg(msg).info(info).loc(&**entry).push();
```
amtep, line 501: Is the &** still needed? You added a From for the Arc<FileEntry>

src/files/fileset.rs
```diff
@@ -694,6 +545,8 @@ impl Fileset {
         }
     }
 
+    /// # Panics
+    /// Will panic if a lock on self.used can not be obtained
```
amtep, line 549: Comment is inaccurate :) See above

src/files/fileset_builder.rs
```diff
@@ -0,0 +1,429 @@
...
+            let inner_path = entry.path().strip_prefix(path).unwrap();
+            if inner_path.starts_with(".git") {
+                continue;
+            }
+            let inner_dir = inner_path.parent().unwrap_or_else(|| Path::new(""));
+            if self.should_replace(inner_dir, kind) {
+                continue;
+            }
+            self.db.files.entry(entry.path().to_path_buf()).or_insert_with(|| {
+                FileEntry::new(inner_path.to_path_buf(), kind, entry.path().to_path_buf())
```
amtep, line 249: Calling `entry.path().to_path_buf()` twice here

src/parse/pdxfile.rs
```diff
@@ -57,42 +58,30 @@ fn parse_pdx(entry: &FileEntry, content: &'static str, memory: &ParserMemory) ->
...
-pub fn parse_pdx_file(
-    entry: &FileEntry,
-    content: String,
-    offset: usize,
-    parser: &ParserMemory,
-) -> Block {
-    let content = leak(content);
-    store_source_file(entry.fullpath().to_path_buf(), &content[offset..]);
-    parse_pdx(entry, &content[offset..], parser)
+pub fn parse_pdx_file(entry: &FileEntry, parser: &ParserMemory) -> Option<Block> {
```
amtep, line 71: I guess this function can be removed and the callers just call parse_pdx

amtep (changes requested):
Mostly comment changes, but the use of PathIdx::UNTRACKED needs to be revisited. The entries that use it can definitely have errors reported. The launcher file via errors emitted by the parser, and the internal strings sometimes have errors that I need to fix. It would be awkward to not get those errors.

PR #271, opened 2025-08-11
Fix Loc ordering

BrknRobot:
Loc order was not stable between runs as `link_index` is non-deterministic.
This implements stable ordering for Loc, giving reports consistent ordering between runs.

To avoid repeated `MACRO_MAP` traversals when comparing expanded log report pointers, `Loc` is split into `Loc` and `LocStack` with `Loc` excluding `link_idx`

### DIFF
--- a/src/config_load.rs
+++ b/src/config_load.rs
@@ -409,7 +409,7 @@ pub fn assert_one_key(assert_key: &str, block: &Block) {
             .map(|(index, key)| PointedMessage {
                 loc: key.into_loc(),
                 length: 1,
-                msg: Some((if index == 0 { "It occurs here" } else { "and here" }).to_owned()),
+                msg: Some((if index == 0 { "It occurs here" } else { "and here" }).into()),
             })
             .collect();
         err(ErrorKey::Config)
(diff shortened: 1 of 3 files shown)


amtep:
So... I don't like the introduction of LocStack. Token and Loc are fundamental to tiger and I don't want to complicate that.

As an alternative for the performance benefit, I think you could set `link_idx` to None when expanding the pointers. It makes sense because the chains are being broken there.

amtep:
Which one is conceptual complexity? :) I want Token and Loc and Block to remain the fundamental concepts used by tiger. I think having two kinds of Loc is going to upset that foundation for very little benefit.

> Other places that would make location comparisons like DbItem insertion should be working with Loc's that have link_idx set to None, or at the very least, Loc's that differ at the top of the stack rather than some number of links down I think, so it's more about explicit expectations than perf for the rest of the codebase. 

Item insertion works with Tokens, so applying this there would mean also splitting TokenWithLocStack from Token... it won't be pretty, and it's an example of how this extra conceptual complexity would work its way through the code base.

PR #293, opened 2025-12-03
Handle Hemegomines & gov flags being in a list & Japanese language

ProZeratul:
I did this mainly to have less false reports in the current build, feel free to take it or not

On the flags changes, I looked at how it was done in the code, might not be the best way, but it seems to work (still not at ease with rust)


amtep:
That's just a confusing ordering :) simp_chinese is in all the games

PR #294, opened 2025-12-21
vic3: 1.12 update

kaiser-chris:
Hi @amtep,

this is a work in progress and just so you can see what I am doing as well as how bad I did it ;)


kaiser-chris:
> So far it looks good :) something went wrong with the datatype munching though. Maybe my fix was wrong

ok gonna take a look
and I have some time before the new year and should get this done until then.

PR #300, opened 2026-01-06
fix(imperator): `start_building_construction` effect was missing

IhateTrains:
It was added in patch 2.0.5:
https://forum.paradoxplaza.com/forum/threads/patch-2-0-5-open-beta.1719724/

> Scope: Province
> 
> start_building_construction = building
> cancel_building_construction = building
> 
> - These two start/cancel the construction of a building in the given province.
> 
> Buildings can be found in common\buildings. A valid example is fortress_building

### DIFF
--- a/src/data/portrait.rs
+++ b/src/data/portrait.rs
@@ -60,7 +60,7 @@ impl DbKind for PortraitModifierGroup {
             }
         }
 
-        if !caller.is_empty() {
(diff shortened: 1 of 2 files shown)


amtep (changes requested):
The real problem here is that `cancel_building_construction` is in there twice :) the second one should be changed to start.

